Issue #5, opened 2019-10-02
Add license info in files


smoia:
Add info about the license in all the main files, as per the end of the GNUv3 license

MinchinWeb:
Do you want the full license text, or just that the file is under GNUv3?

smoia:
It should be following all the specifications [from line 623 to the last line of the licence](https://github.com/smoia/phys2bids/blob/c7976275787f6c6ad637dfb5dbc081de05ccefd9/LICENSE#L623)

PR #14, opened 2019-10-05
Update physiobids.py

CodeLongAndProsper90:


smoia:
Sorry, I'll try to explain better what I wrote in #5 :
It's not the lines that should be copy-pasted in the script.
They should be read and what is inside should be done: Adding just the part relevant to the code in the script at its beginning, and then add a print command at the beginning of the interaction with the user.

I'm going to assing a member of the devteam to this. 
Thank you, and sorry I can't consider this PR valid.

PR #34, opened 2019-11-20
Requirements

smoia:
Updated requirements with bioread version

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-bioread==1.0.4
+bioread>=1.0.4
 numpy>=1.9.3
 pandas>=0.10
 matplotlib>=3.1.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ provides =
 [options]
 python_requires = >=3.6.1
 install_requires =
-    bioread
+    bioread ==1.0.4
     numpy >=1.9.3
     pandas >=0.10
     matplotlib >=3.1.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ provides =
 [options]
 python_requires = >=3.6.1
 install_requires =
-    bioread ==1.0.4
+    bioread >=1.0.4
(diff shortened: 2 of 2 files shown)


smoia:
This is an easy PR: I just checked the version of bioread that we're using and added it to the requirements for setup and pip.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-bioread
+bioread==1.0.4
```
rmarkello: Any reason we're pinning this as *exactly* equal rather than a `>=`?

smoia: Compatibility, mainly.
As an alternative, we can change it to bioread >= 1.0.4 < 2.0.0

rmarkello: Ooh, what about the code in `bioread >= 2.0.0` breaks the current pipeline? I generally don't like having `<` in dependencies, but if it's unavoidable then so be it.

(To answer your question, though: yes, I would definitely prefer `bioread>=1.0.4,<2.0.0` to what we have now!)

smoia: I don't know, bioread 2.0.0 hasn't been written yet! :smile: 
However if the dev uses semantic versioning, we can expect it to break, no?
Actually, the same goes for numpy  - we should be able to remove pandas dependencies!

rmarkello: Right, but you don't want to preempt these things, generally speaking. If any of the dependencies releases a breaking change then tests will begin to fail on the `phys2bids` master branch and you can release a new version of `phys2bids` with updates to address the broken changes.

For now I would say just do `>=` for all these dependencies so we have a good starting point (i.e., we know that, as of right now, people will have the minimum required versions of the packages so that `phys2bids` works). We can address any breaking changes as they arise in the future.

smoia: I see. Ok, no problem!

rmarkello:
Looks good to me! Should we go ahead and merge despite the failing tests, since we're going to fix those separately?

smoia:
Indeed!

PR #35, opened 2019-11-20
Changed name of files in test folder

smoia:


smoia:
This is another easy one: i changed the name of the files in testing to follow the spec in gitter. You can check it's right and if it is, merge it!

smoia:
I think it's less confusing if we just flag those files that don't have a trigger inside with the "no_trigger" flag.

Issue #23, opened 2019-11-14
Internal I/O object/schema


smoia:
So far, the code was set to adapt to whatever input was provided and process it to an output on a case to case base.
This is **not** a good approach, not only due to the flexibility we need to offer, but also from a coding perspective.

It's much better to create an object ("Output object"), that contains not only the data but a schema that can be used to:
a. output information about the file that can be used by the user to properly set the input.
b. get the name of the file and the info for the json file.

In this way, the main code would operate on this object, while another family of scripts (interfaces?) would work on populating the object schema based on the type of input and the information embedded in it.

You can think about the new object as composed of a numpy matrix (data), some lists (values that are defined channel by channel) and some variables that contain general information.

This said, what should be the adopted schema (i.e., what information do we need to carry around)?
We know that BIDS has certain requirements, so those have to be in the schema.
However, we can expand the schema thinking about what we'd like to have there for general purposes.

RayStick:
Issue #36 may be relevant for this discussion

Issue #1, opened 2019-09-16
Write docstrings


smoia:
Add numpy docstrings to functions

smoia:
Most of the scripts we have require docstrings, with the exception of those in `heur` folder.
Just to be clear, docstrings are a multiline comment at the beginning of each function that briefly explain what the function is doing, and reports Inputs, Outputs and (after @eurunuela 's comment in #38) outcomes (when the function doesn't have a `return`, but produces something like a file or a print onscreen).
At the moment, `physio_obj.py` has complete docstrings, `phys2bids.py` has complete docstrings with the only exception of method `main()`; `utils.py` has stubs to complete with info on I/O, `viz.py` lacks docstrings.

It's a good first issue to start reading the code and fill in the characteristics of the functions. Python knowledge is not necessarily required.

smoia:
I would still invite someone that would like to start approaching phys2bids code / python3 to complete this issue. Especially in improving `utils.py` docstrings - most of the functions there are quite trivial, and a valid PR doesn't have to cover everything.
Can I suggest @kristinazvolanek , @RayStick or @AyyagariA  for this? 
I can help out remotely! Drop me a mesage somewhere if you're interested.

RayStick:
I am happy to help with docstrings once I have used the code a bit more. Personally, I think it is hard to write docstrings if you haven't written or used the code. So at the very least, we need to use the code, if we are not writing it.

eurunuela:
Given that this issue was labeled as urgent, how are we doing with this?

RayStick:
It seems that @smoia made good progress on this?

smoia:
This should have been closed automatically by #80 . Closing it now.

PR #38, opened 2019-11-21
Restructure

### LINKED ISSUE #10: Split input in different outputs based on the sample frequency of the channels
BIDS wants each *_physio.tsv.gz file to have the same frequency internally.
When a input with different frequencies is given, the program should be able to export it in different outputs.

smoia:
Resolves #10 .
Address #23 , #1 .
After sharing some thoughts with @rmarkello , I'm reformatting the code so that we have an object to populate within interfaces scripts - each one to call based on the input.


smoia:
Still working on `phys2bids.py`, but the rest should be ready.

smoia:
There's still docstrings to add all around, and I'm sure the code can be refactored better, but for the moment let's see if it's ok!

smoia:
As soon as this is merged, I'm moving the repo to the organisation!

phys2bids/phys2bids.py
```diff
@@ -119,160 +103,111 @@ def _main(argv=None):
...
+        # The next few lines create a dictionary of different blueprint_input
+        # objects, one for each unique frequency in phys_in
+        uniq_freq_list = set(phys_in.freq)
+        print(f'Found {len(uniq_freq_list)} unique frequencies.')
+        phys_out = {}
+        for uniq_freq in uniq_freq_list:
+            phys_out[uniq_freq] = deepcopy(phys_in)
+            for i in [i for i, x in enumerate(reversed(phys_in.freq))
+                      if x != uniq_freq]:
+                phys_out[uniq_freq].delete_at_index(phys_in.ch_amount-i-1)
```
eurunuela: Same as before, this is an odd way to handle the removal of other frequencies imo. The following is much simpler:

```
for i in reversed(phys_in.freq):
    if i != uniq_freq:
        phys_out[uniq_freq].delete_at_index(phys_in.ch_amount-i-1)
```

phys2bids/phys2bids.py
```diff
@@ -119,160 +103,111 @@ def _main(argv=None):
...
+
+        for uniq_freq in uniq_freq_list:
+            phys_out[uniq_freq] = blueprint_output.init_from_blueprint(phys_out[uniq_freq])
+
+        output_amount = len(uniq_freq_list)
+        if output_amount > 1:
+            print(f'Found {output_amount} different frequencies in input!\n'
+                  f'Consequently, preparing {output_amount} of output files')
```
eurunuela: This is a matter of tastes but imo the word _Consequently_ should not be there. It just sounds weird to me from an end-user pov.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+    checks that the list contains list_type.
+    Input
+    -----
+    var:
+        Variable to be checked.
+    var_type: type
+        Type the variable is assumed to be.
+    list_type: type
+        As type.
+
```
eurunuela: You're missing the `return_var` argument in the docstring.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+    Checks that the var has the same dimension of the data
+    If it's not the case, fill in the var or removes exceding var entry.
+    Input
+    -----
+    var:
+        Variable to be checked.
+    data_size: int
+        Size of data of interest.
+    token:
+        What to be used in case of completion.
```
eurunuela: This is not descriptive enough.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+def has_size(var, data_size, token):
+    """
+    Checks that the var has the same dimension of the data
+    If it's not the case, fill in the var or removes exceding var entry.
+    Input
+    -----
+    var:
+        Variable to be checked.
```
eurunuela: What is the type to expect? Is it a numpy array?

eurunuela: Also, you should indicate that the shape of the array (if it's an array) is `data_size x 1`.

eurunuela: I see that `var` is always a list.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+    def rename_channels(cls, new_names, ch_trigger=None):
+        """
+        Renames the channels.
+
+        Input
+        -----
+        new_names: list of str
+            New names for channels.
+        ch_trigger:
+            Number of the channel containing the trigger.
```
eurunuela: You're missing `cls`.

eurunuela: Actually, if by `cls` you mean the object, the input should be `self`.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def return_index(cls, idx):
+        """
+        Returns the proper list entry of all the
+        properties of the object, given an index.
+
+        Input
+        -----
+        idx: int
+            Index of elements to return
```
eurunuela: Same as before. You're missing `cls`.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+                          'tr or with a more liberal thr')
+
+            else:
+                print('Found just the right amount of tps!')
+
+        else:
+            print('Cannot check the number of tps')
+
+        cls.timeseries[0] -= time_offset
+        cls.num_tps_found = num_tps_found
```
eurunuela: You should be using `self` and not `cls`.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+
+        Input
+        -----
+        idx: int or range
+            Index of elements to delete from all lists
+         """
+        del(cls.timeseries[idx])
+        del(cls.freq[idx])
+        del(cls.ch_name[idx])
+        del(cls.units[idx])
```
eurunuela: You should use `self` and not `cls`.

phys2bids/physio_obj.py
eurunuela: You should actually be using `self`.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+            Threshold to be used to detect trigger points.
+            Default is 2.5
+        num_tps_expected: int
+            Number of expected triggers (num of TRs in fMRI)
+        tr: float
+            The Repetition Time of the fMRI data.
+        """
+        print('Counting trigger points')
+        trigger_deriv = np.diff(cls.timeseries[1])
+        tps = trigger_deriv > thr
```
eurunuela: You should probably rename `tps` to something more meaningful.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+    @classmethod
+    def init_from_blueprint(cls, blueprint):
+        """
+        Method to populate the output blueprint using blueprint_input.
+
+        Input
+        -----
+        blueprint: :obj: blueprint_input
+            The input blueprint object
+        """
```
eurunuela: You should comment on the output as well.

phys2bids/physio_obj.py
eurunuela: Also, you should comment on the output.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def delete_at_index(cls, idx):
+        """
+        Returns all the proper list entry of the
+        properties of the object, given an index.
+
+        Input
+        -----
+        idx: int or range
+            Index of elements to delete from all lists
```
eurunuela: You should comment on the output.

phys2bids/phys2bids.py
```diff
@@ -107,7 +88,10 @@ def use_heuristic(heur_file, sub, ses, filename, outdir):
     heur = utils.load_heuristic(heur_file)
     name = heur.heur(filename[:-4], name)
 
-    heurpath = fldr + '/' + name + '_physio'
+    if record_label:
+        recording = f'_recording-{record_label}'
+
+    heurpath = fldr + '/' + name + recording + '_physio'
```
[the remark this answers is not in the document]
smoia: I opened a "Good First Issue" #40 to deal with that. I think it's good if someone with less experience wants to get more used to the code! Can we leave it out of this PR?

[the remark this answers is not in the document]
eurunuela: Sure. This is not critical for the PR.

phys2bids/phys2bids.py
```diff
@@ -119,160 +103,111 @@ def _main(argv=None):
...
+    outfile = os.path.join(options.outdir, os.path.basename(options.filename[:-4]))
 
-    # Read infos from file
+    # Read file!
     if ftype == 'acq':
-        from bioread import read_file
-
-        data = read_file(infile).channels
-        print_info_acq(options.filename, data)
+        from phys2bids.interfaces.acq import populate_phys_input
```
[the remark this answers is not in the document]
smoia: Mmm I kinda disagree with this.
I think a better way to deal with this is if each interfaces script contains a "populate_phys_input" function that we import based on which file we're dealing with. I know it's not massive code underperforming, but I'd like to hear what the other reviewers think about it.

phys2bids/physio_obj.py
smoia: var can actually be anything, at the moment it's either a list or a numpy array. I can add that. The shape changes based on the input

phys2bids/physio_obj.py
smoia: Not according to pep8, at least [here](https://www.python.org/dev/peps/pep-0008/#id46)

phys2bids/phys2bids.py
```diff
@@ -119,160 +103,111 @@ def _main(argv=None):
...
     elif ftype == 'txt':
-        header = print_info_txt(options.filename)
+        raise Exception('txt not yet supported')
+    else:
+        raise Exception('This shouldn\'t happen, check out the last few '
+                        'lines of code')
```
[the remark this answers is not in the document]
smoia: This is a developer only possible code break - line [108](https://github.com/smoia/phys2bids/blob/cfc333e7d85666aeec25b58a6983b506221ecd78/phys2bids/phys2bids.py#L108) should be already raising an exception of currently not supported.
If I understood how things work, to make a debug warning we need a logger. There's an open issue for that in #28 .
What if we put a comment to add a logger here, and I transform that exception in "currently unsupported" filetype?

phys2bids/physio_obj.py
smoia: Should I write "any" for types?

phys2bids/physio_obj.py
smoia: Should I specify "self" in the docstrings? I wouldn't like the user to get confused by it.

phys2bids/physio_obj.py
smoia: It doesn't have an output per se, but an outcome.
Should we write outcomes?

smoia:
Where should the code be more commented? In general, as docstrings, ...?

phys2bids/phys2bids.py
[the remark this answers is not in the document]
eurunuela: Sure!

phys2bids/physio_obj.py
eurunuela: I'd say we should always comment on whatever we return.

eurunuela:
> Where should the code be more commented? In general, as docstrings, ...?

Either in general or as docstrings, as long as it makes it easier to understand what's going on in each part of the code.

phys2bids/physio_obj.py
smoia: We don't return anything!

phys2bids/physio_obj.py
eurunuela: I would write `list or numpy array` so then we now what the can be. Specially because you're appending stuff to `var` and this could be a source of conflicts.

phys2bids/physio_obj.py
eurunuela: Maybe @rmarkello will know better but the way I've always seen this is `self`.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def return_index(cls, idx):
+        """
+        Returns all the proper list entry of the
+        properties of the object, given an index.
+
+        Input
+        -----
+        idx: int
+            Index of elements to return
```
eurunuela: Add comment on output.

phys2bids/physio_obj.py
eurunuela: I'm sorry, this was meant for line 289.

phys2bids/phys2bids.py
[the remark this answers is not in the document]
smoia: @vinferrer @rmarkello

phys2bids/physio_obj.py
smoia: In any case, it's updated with "Outcomes"

smoia:
> I would write list or numpy array so then we now what the can be. Specially because you're appending
> stuff to var and this could be a source of conflicts.
The type of variable can be literally anything. We're using it in the code for lists, floats, ints, and strings at least. The only type it is not used for at the moment is tuples (we don't have many in the code) and numpy arrays. However, to avoid the possible conflict, the function `is_valid` is called on the `token` variable.
I added "any type", I don't think it makes sense to narrow it further in the docs.

eurunuela:
> > I would write list or numpy array so then we now what the can be. Especially because you're appending
> > stuff to var and this could be a source of conflicts.
> 
> The type of variable can be anything. We're using it in the code for lists, floats, ints, and strings at least. The only type it is not used for at the moment is tuples (we don't have many in the code) and numpy arrays. However, to avoid the possible conflict, the function `is_valid` is called on the `token` variable.
> I added "any type", I don't think it makes sense to narrow it further in the docs.

Alright!

phys2bids/phys2bids.py
```diff
@@ -116,163 +170,126 @@ def use_heuristic(heur_file, sub, ses, filename, outdir):
...
+        # Create a blueprint_output object for each unique frequency found.
+        # Populate it with the corresponding blueprint input and replace it
+        # in the dictionary.
+        for uniq_freq in uniq_freq_list:
+            phys_out[uniq_freq] = blueprint_output.init_from_blueprint(phys_out[uniq_freq])
+
+        output_amount = len(uniq_freq_list)
+        if output_amount > 1:
+            print(f'Found {output_amount} different frequencies in input!\n'
+                  f'Preparing {output_amount} output files')
```
rmarkello: I think we already print something to similar effect on L241? Do you think we could get rid of this bit if that's the case?

phys2bids/cli/run.py
```diff
@@ -89,20 +89,18 @@ def _get_parser():
...
-                          dest='table_header',
+    optional.add_argument('-chnames', '--channel-names',
+                          dest='ch_name',
                           nargs='*',
                           type=str,
                           help='Columns header (for json file).',
-                          # #!# Has to go to empty list
-                          default=['time', 'respiratory_chest', 'trigger',
-                                   'cardiac', 'respiratory_CO2', 'respiratory_O2'])
+                          default=[''])
```
rmarkello: Having the default as `['']` will mean that `options.ch_name` in the` phys2bids.py` module will always evaluate to True. I think we should just leave the default off (which means, if not provided, it will default to None), and then check for equivalency to NoneType in `phys2bids.py`. I'll make a comment in the relevant spot!

phys2bids/phys2bids.py
```diff
@@ -116,163 +170,126 @@ def use_heuristic(heur_file, sub, ses, filename, outdir):
...
+
+        # The next few lines remove the undesired channels from phys_in.
+        if options.chsel:
+            print('Dropping unselected channels')
+            for i in reversed(range(0, phys_in.ch_amout)):
+                if i not in options.chsel:
+                    phys_in.delete_at_index(i)
+
+        # If requested, change channel names.
+        if options.ch_name:
```
rmarkello: I think this should be updated to:

```suggestion
        if options.ch_name is not None:
```

if you make the requested change in `phys2bids/cli/run.py`!

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+    def __init__(self, timeseries, freq, ch_name, units):
+        self.timeseries = is_valid(timeseries, list, list_type=np.ndarray)
+        self.ch_amount = len(self.timeseries)
+        self.freq = has_size(is_valid(freq, list,
+                                      list_type=(int, float)),
+                             self.ch_amount, 0)
+        self.ch_name = has_size(ch_name, self.ch_amount, 'unknown')
+        self.units = has_size(units, self.ch_amount, '[]')
+
+    def rename_channels(cls, new_names, ch_trigger=None):
```
rmarkello: This is a method for an instance of `blueprint_input`, not a class method, correct? If so then I think it would be great for this (and all subsequent `cls` params) to be revised to `self`! While this will still function as-written, it's very confusing to read!

```suggestion
    def rename_channels(self, new_names, ch_trigger=None):
```

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+        Output
+        ------
+        out: tuple
+            Tuple containing the proper list entry of all the
+            properties of the object with index `idx`
+        """
+        return (cls.timeseries[idx], cls.ch_amount, cls.freq[idx],
+                cls.ch_name[idx], cls.units[idx])
+
+    def delete_at_index(cls, idx):
```
rmarkello:
```suggestion
    def delete_at_index(self, idx):
```

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+        cls:
+            In all the property that are lists, the element correspondent to
+            `idx` gets deleted
+         """
+        del(cls.timeseries[idx])
+        del(cls.freq[idx])
+        del(cls.ch_name[idx])
+        del(cls.units[idx])
+
+    def check_trigger_amount(cls, thr=2.5, num_timepoints_expected=0, tr=0):
```
rmarkello:
```suggestion
    def check_trigger_amount(self, thr=2.5, num_timepoints_expected=0, tr=0):
```

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+    """
+    def __init__(self, timeseries, freq, ch_name, units, start_time):
+        self.timeseries = is_valid(timeseries, np.ndarray)
+        self.ch_amount = self.timeseries.shape[0]
+        self.freq = is_valid(freq, (int, float))
+        self.ch_name = has_size(ch_name, self.ch_amount, 'unknown')
+        self.units = has_size(units, self.ch_amount, '[]')
+        self.start_time = start_time
+
+    def return_index(cls, idx):
```
rmarkello:
```suggestion
    def return_index(self, idx):
```

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+        Output
+        ------
+        out: tuple
+            Tuple containing the proper list entry of all the
+            properties of the object with index `idx`
+        """
+        return (cls.timeseries[idx], cls.ch_amount, cls.freq,
+                cls.ch_name[idx], cls.units[idx], cls.start_time)
+
+    def delete_at_index(cls, idx):
```
rmarkello:
```suggestion
    def delete_at_index(self, idx):
```

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+        cls:
+            In all the property that are lists, the element correspondent to
+            `idx` gets deleted
+        """
+        del(cls.timeseries[idx])
+        del(cls.ch_name[idx])
+        del(cls.units[idx])
+
+    @classmethod
+    def init_from_blueprint(cls, blueprint):
```
rmarkello: :+1: on using `cls` here!

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+    - units[0] = 's'
+    - Actual number of channels (ANC) +1 <= ch_amount <= ANC +2
+    """
+    def __init__(self, timeseries, freq, ch_name, units):
+        self.timeseries = is_valid(timeseries, list, list_type=np.ndarray)
+        self.ch_amount = len(self.timeseries)
+        self.freq = has_size(is_valid(freq, list,
+                                      list_type=(int, float)),
+                             self.ch_amount, 0)
+        self.ch_name = has_size(ch_name, self.ch_amount, 'unknown')
```
rmarkello: I believe that `ch_name` here is generally a list, in which case this will fail with:

```python-traceback
~/phys2bids/physio_obj.py in is_valid(var, var_type, list_type, return_var)
     32     """
     33     if not isinstance(var, var_type):
---> 34         raise AttributeError(f'The given variable is not a {var_type}')
     35 
     36     if var_type is list and list_type is not None:

AttributeError: The given variable is not a <class 'list'>
```

Either the final parameter here should be ['unknown'] or we need to update the `has_size` and `is_valid` functions to better handle these different parameters combos we're hitting!

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+        self.units = has_size(units, self.ch_amount, '[]')
```
rmarkello: This is also likely to fail for the same reason as above (unless `units` and `self.ch_amount` are equal).

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+        Outcome
+        -------
+        cls:
+            In all the property that are lists, the element correspondent to
+            `idx` gets deleted
+         """
+        del(cls.timeseries[idx])
+        del(cls.freq[idx])
+        del(cls.ch_name[idx])
+        del(cls.units[idx])
```
rmarkello: Should we also be subtracting one from `self.ch_amount`?

phys2bids/interfaces/acq.py
```diff
@@ -0,0 +1,33 @@
...
+
+"""
+phys2bids interface for acqknowledge files.
+"""
+from bioread import read_file
+
+from phys2bids.physio_obj import blueprint_input
+
+
+def populate_phys_input(filename, chtrig):
```
rmarkello: So `chtrig` is not really an optional parameter, here? It's possible that a user would not have supplied it since it's currently listed as "optional" in the argument parser; if it's necessary for functionality though then we should change that!

phys2bids/physio_obj.py
rmarkello: The exact code that is being passed (and failing) is:

```python
self.ch_name = has_size(['time', 'trigger'], 7, 'unknown')
```

if you want to recreate the error locally. I was using the `.acq` file in the `tests/data` directory!

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+                print('Found just the right amount of timepoints!')
+
+        else:
+            print('Cannot check the number of timepoints')
+
+        cls.timeseries[0] -= time_offset
+        cls.num_timepoints_found = num_timepoints_found
+
+
+class blueprint_output():
```
rmarkello:
```suggestion
class BlueprintOutput:
```

Again, capital case is recommended for classes!

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+    Methods
+    -------
+    return_index:
+        Returns the proper list entry of all the
+        properties of the object, given an index.
+    delete_at_index:
+        Returns all the proper list entry of the
+        properties of the object, given an index.
+    init_from_blueprint:
+        method to populate from input blueprint instead of init
```
rmarkello: I'm not sure I'm getting what benefits this has over `blueprint_input`. The main differences (from what I can tell!) are that `timeseries` is required to be an array instead of a list of arrays, and this doesn't have a few of the methods that the `blueprint_input` does (e.g., `.check_trigger_amount()` and `.rename_channels()`). Could you clarify what purpose this class is serving over-and-above `blueprint_input`?

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+    """
+    def __init__(self, timeseries, freq, ch_name, units, start_time):
+        self.timeseries = is_valid(timeseries, np.ndarray)
+        self.ch_amount = self.timeseries.shape[0]
+        self.freq = is_valid(freq, (int, float))
+        self.ch_name = has_size(ch_name, self.ch_amount, 'unknown')
+        self.units = has_size(units, self.ch_amount, '[]')
```
rmarkello: Same errors as with `blueprint_input`, above.

phys2bids/phys2bids.py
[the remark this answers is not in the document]
smoia: I don't think we'll ever need different libraries to handle txt files though. Those should be imported with numpy or pandas!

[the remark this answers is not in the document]
rmarkello: Okay! As long as there aren't any other proprietary software used to collect physio data with their own, weird formats then that sounds good to me :+1:

phys2bids/phys2bids.py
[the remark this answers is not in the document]
smoia: Right, that's "Good First Issue" #40 . @eurunuela had similar concerns.
Can we keep it in that issue and go on with this PR separately? Otherwise I'll do all of that one.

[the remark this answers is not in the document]
rmarkello: Sorry, wasn't referencing the issues when looking at this. That's fine!

phys2bids/utils.py
```diff
@@ -4,19 +4,49 @@
...
-        file = file + ext
+def check_input_ext(filename, ext):
+    """
+    Checks that the given file has the given extension
+    """
+    if filename[-len(ext):] != ext:
+        filename = filename + ext
 
-    return file
+    return filename
```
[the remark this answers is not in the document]
smoia: This will fail in the case of .nii.gz though (foreseeing application). We can use both?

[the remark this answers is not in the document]
smoia: Actually this will fail in any possible `.*.gz`, like `.tsv.gz`.

I'd like a safe solution for such cases.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,403 @@
...
+            del(new_names[new_names.index['trigger']])
+        elif ch_trigger:
+            del(new_names[ch_trigger])
+
+        new_names = ['time', 'trigger'] + new_names
+
+        cls.ch_name = has_size(is_valid(new_names, list, list_type=str),
+                               cls.ch_amount, 'unknown')
+
+    def return_index(cls, idx):
```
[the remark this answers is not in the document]
smoia: Is this related to what @eurunuela was pointing out before?

phys2bids/phys2bids.py
```diff
@@ -116,163 +170,126 @@ def use_heuristic(heur_file, sub, ses, filename, outdir):
...
+        from phys2bids.interfaces.acq import populate_phys_input
     elif ftype == 'txt':
-        header = print_info_txt(options.filename)
+        raise Exception('txt not yet supported')
+    else:
+        # #!# We should add a logger here.
+        raise Exception('Currently unsupported file type.')
+
+    print('Reading the file')
+    phys_in = populate_phys_input(infile, options.chtrig)
```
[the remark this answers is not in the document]
smoia: For the moment, the program should terminate.
However, we need to implement the option of not having a trigger channel - it's not just bad practice, it's also some tools like Respiract or similar that do not collect triggers from the MR. Not in this PR.

phys2bids/cli/run.py
smoia: Doesn't it interfere with the check that the parser is doing on the type of input we expect?

phys2bids/interfaces/acq.py
smoia: Let's do the change for now to a required input.

smoia: Sorry, this is actually optional but has a valid default (`0`). Still needed as a required variable or this way is okish?

phys2bids/physio_obj.py
smoia: So here's the thing. For latter uses (e.g. if we go on with preprocessing or data analysis), I'd say numpy arrays are a bit easier to work with, rather than list of numpy arrays.
However, as the input might have different frequencies, we need lists at the beginning.
Also, in one frequencies are a list and in the other it's a float, and we need to really check this and be sure it's respected.

What would be a solution otherwise?

phys2bids/physio_obj.py
smoia: Sorry, bug that wasn't failing on my part due to a change to my variables I did while debugging.
Corrected now though.

phys2bids/interfaces/acq.py
rmarkello: Gotcha! Okay, let's address the whole `chtrig` exists sometimes but not other times in a separate issue :+1:

phys2bids/physio_obj.py
eurunuela: That was my point. We should make sure that it works with different types if that's what we're going for. I would leave it for a future PR anyway.

phys2bids/phys2bids.py
```diff
@@ -107,7 +158,10 @@ def use_heuristic(heur_file, sub, ses, filename, outdir):
     heur = utils.load_heuristic(heur_file)
     name = heur.heur(filename[:-4], name)
 
-    heurpath = fldr + '/' + name + '_physio'
+    if record_label:
+        recording = f'_recording-{record_label}'
+
+    heurpath = fldr + '/' + name + recording + '_physio'
```
vinferrer: Hi guys, 

Leaving apart the "Good first issue" #40 apart. the variable ``recording`` is defined inside the if and therefore if not defined before as an empty string or something if ``record_label`` is left as default, the function will give and error for sure. I think either you put ``record_label`` as a required argument so there is no problem and then delete that if statement or define ``recording`` before the conditional statement.

phys2bids/phys2bids.py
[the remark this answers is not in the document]
vinferrer: I agree with @smoia and @rmarkello. If this ``populate_phys_input`` is only use with ``.acq`` files it should only be load in this case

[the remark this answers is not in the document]
vinferrer: Actually, thinking about this it is likely that in the future we should create a module inside phys2bids.interfaces for each input format, this way we can name the reading functions the same way or give the same alias in the main function. What do you think?

phys2bids/phys2bids.py
```diff
@@ -107,7 +158,10 @@ def use_heuristic(heur_file, sub, ses, filename, outdir):
     heur = utils.load_heuristic(heur_file)
     name = heur.heur(filename[:-4], name)
```
vinferrer:
```suggestion

recording = ''
```

phys2bids/phys2bids.py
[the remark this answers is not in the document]
smoia: Yep, that's the idea @vinferrer .
I also like your idea of creating methods with different names in different modules, but importing them with the same name in the main workflow.

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,400 @@
...
+        Variable to be checked (same as input).
+    """
+    if not isinstance(var, var_type):
+        raise AttributeError(f'The given variable is not a {var_type}')
+
+    if var_type is list and list_type is not None:
+        for element in var:
+            _ = is_valid(element, list_type, return_var=False)
+
+    return var
```
vinferrer: Shouldn't this be in ``utils.py`` why do you put it in the definition of the ``physio_obj.py``?

phys2bids/physio_obj.py
```diff
@@ -0,0 +1,400 @@
...
+    def init_from_blueprint(cls, blueprint):
+        """
+        Method to populate the output blueprint using BlueprintInput.
+
+        Input
+        -----
+        cls: :obj: `BlueprintOutput`
+            The object on which to operate
+        blueprint: :obj: `BlueprintInput`
+            The input blueprint object
```
vinferrer: This is actually not true, the input here is ``BlueprintInput[selected_frequency]`` no?

phys2bids/physio_obj.py
vinferrer: Oh i realize now that is specific for the object, sry for the disturbance

phys2bids/utils.py
```diff
@@ -27,12 +61,23 @@ def path_exists_or_make_it(fldr):
...
+
+def print_info(filename, phys_object):
+    """
+    Print the info of the input files, using BlueprintInput object
+    """
+    print(f'File {filename} contains:\n')
+
+    for ch in range(2, phys_object.ch_amount):
+        print(f'{(ch-2):02d}. {phys_object.ch_name[ch]};'
+              f' sampled at {phys_object.freq[ch]} Hz')
```
vinferrer: Shouldn't this be a method inside ``BlueprintInput``?

phys2bids/physio_obj.py
smoia: I don't think so. In `phys2bids.py`, `phys_out` is a dictionary that contains a bunch of `BlueprintOutput` objects. Here, it's the class itself.

phys2bids/utils.py
smoia: We can move it there, yes.

PR #49, opened 2019-11-28
[TST/WIP] Adds unit tests

eurunuela:


eurunuela:
In the end, I thought it would be much better to have a PR with all the unit tests, given that it won't take me much to have them ready.

PR #54, opened 2019-11-29
Adding txt support for labchart 

### LINKED ISSUE #48: Add txt support
Similar to #41 , but at the very least we should be able to process **a txt file with a header coming from labchart**.
On OSF there's [this one](https://mfr.de-1.osf.io/render?url=https://osf.io/b3cq2/?action=download%26mode=render).
Bonus: be able to run the txt file in this repo (folder tests/data), that comes from acqknowledge.

This requires setting up a new script in `interfaces`, similar to the existing `acq.py`.

vinferrer:
close #48 

Hi guys, I implemented txt support for labchart files, could you review my code? no rush.

Cheers,
Vicente

### DIFF
--- a/phys2bids/interfaces/txt.py
+++ b/phys2bids/interfaces/txt.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 """
-phys2bids interface for acqknowledge files.
+phys2bids interface for txt files.
 """
 from bioread import read_file
 
--- a/phys2bids/interfaces/txt.py
+++ b/phys2bids/interfaces/txt.py
@@ -12,13 +12,13 @@ def populate_phys_input(filename, chtrig):
     """
     Populate object phys_input
     for now this works only with labchart files
-     Input (Properties)
+        Input (Properties)
     ------------------
     filename: str 
         path to the txt labchart file
     chtrig : int
-       index of trigger channel 
-    
+        index of trigger channel
+
     Output
     ------------------
     BlueprintInput populated object for more see BlueprintInput docs  
@@ -66,7 +66,7 @@ def populate_phys_input(filename, chtrig):
         units.append(item.split(' ')[1])
     # get names
     orig_names=header[4][1:]
-    names = ['time',orig_names[chtrig]]
+    names = ['time','trigger']
     orig_names.pop(chtrig)
     names=names+orig_names
     # get channels 
--- a/phys2bids/interfaces/txt.py
+++ b/phys2bids/interfaces/txt.py
@@ -4,7 +4,6 @@
 """
 phys2bids interface for txt files.
 """
-from bioread import read_file
 
 from phys2bids.physio_obj import BlueprintInput
 
--- a/phys2bids/interfaces/txt.py
+++ b/phys2bids/interfaces/txt.py
@@ -11,7 +11,7 @@ import numpy as np
 def populate_phys_input(filename, chtrig):
     """
     Populate object phys_input
-    for now we this works with labchart files
+    for now this works only with labchart files
     """
     header = []
     channel_list = []
--- a/phys2bids/interfaces/txt.py
+++ b/phys2bids/interfaces/txt.py
@@ -21,7 +21,7 @@ def populate_phys_input(filename, chtrig):
 
     Output
     ------------------
-    BlueprintInput populated object for more see BlueprintInput docs  
+    BlueprintInput object for more see BlueprintInput docs  
     """
 
     header = []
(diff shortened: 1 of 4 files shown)


phys2bids/interfaces/txt.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""
+phys2bids interface for txt files.
+"""
+from bioread import read_file
```
smoia:
```suggestion

```

phys2bids/interfaces/txt.py
```diff
@@ -0,0 +1,64 @@
...
+
+from phys2bids.physio_obj import BlueprintInput
+
+import numpy as np
+# '/home/vicente/phys2bids/phys2bids/tests/data/Test_multiscan100Hz_trig_samefreq_header.txt'
```
smoia:
```suggestion

```

phys2bids/interfaces/txt.py
```diff
@@ -0,0 +1,64 @@
...
+def populate_phys_input(filename, chtrig):
+    """
+    Populate object phys_input
+    for now we this works with labchart files
```
smoia: Can you reformulate the sentence please?

phys2bids/interfaces/txt.py
```diff
@@ -0,0 +1,64 @@
...
+    """
```
smoia: It would be great if you could add Inputs and Outputs to complete the docstring (see `phys_obj.py`).
Then we can use it for `acq.py`.

phys2bids/interfaces/txt.py
```diff
@@ -0,0 +1,64 @@
...
+                float(line[0])
+            except ValueError:
+                header.append(line)
+                continue 
+            line = [float(i) for i in line]
+            channel_list.append(line)
+    # get frequency 
+    interval = header[0][1].split(" ")
+    if interval[-1] not in ['hr', 'min', 's','ms','µs']:
+        raise AttributeError(f'Interval is not in a valid LabChart time unit, this probably means your file is not in labchart format')
```
smoia:
```suggestion
        raise AttributeError(f'Interval unit "{interval[-1]}" is not in a valid LabChart time unit, '
                             'this probably means your file is not in labchart format')
```

phys2bids/interfaces/txt.py
```diff
@@ -0,0 +1,64 @@
...
+
+    if interval[-1] != 's':
+        print(f'interval is not in s converting')
```
smoia: f-string is not necessary here! You can leave it as a normal string.
```suggestion
        print('Interval is not in seconds. Converting its value.')
```

phys2bids/interfaces/txt.py
```diff
@@ -0,0 +1,64 @@
...
+    else:
+        interval[0] = float(interval[0])
+    # get units
+    range_list = header[5][1:]
+    units = []
+    for item in range_list:
+        units.append(item.split(' ')[1])
+    # get names
+    header[4][1:]
+    names = ['time']+header[4][1:]
```
smoia: This is great, because our timeseries will start with a "time" list.
However, internally the class requires the "trigger" list/channel to be the second element in the timeseries list.
Can you add this feature? Otherwise there's an incompatibility (or we risk an incompatibility) later on.
That's also why you need to feed `chtrig` to this function!

smoia: You can add the check here for names and after you assign `timeseries` for the element itself.

phys2bids/interfaces/txt.py
```diff
@@ -0,0 +1,64 @@
...
+# -*- coding: utf-8 -*-
+
+"""
+phys2bids interface for txt files.
+"""
+from bioread import read_file
+
+from phys2bids.physio_obj import BlueprintInput
+
+import numpy as np
```
smoia: Can you invert import of numpy and import of phys2bids?
It's more similar to the rest of the code (external mod first, then internal mod)

vinferrer:
I have done the changes, review the if you want and accept them

phys2bids/interfaces/txt.py
smoia: I know this is picky, but can you check the indentation of the docstring?

phys2bids/interfaces/txt.py
```diff
@@ -54,10 +54,16 @@ def populate_phys_input(filename, chtrig):
     for item in range_list:
         units.append(item.split(' ')[1])
     # get names
-    header[4][1:]
-    names = ['time']+header[4][1:]
+    orig_names=header[4][1:]
+    names = ['time',orig_names[chtrig]]
```
smoia: Can you call it literally `'trigger'`?

vinferrer: Done and also revised the docstring identation

PR #55, opened 2019-11-30
[DOC] Adds contributing guidelines

eurunuela:
This is the first step into having a contributing guideline and adding the contributors bot.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,313 @@
+# Contributing to `phys2bids`
+
+Welcome to the `phys2bids` repository! We're excited you're here and want to contribute.
+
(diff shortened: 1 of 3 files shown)


smoia:
Thanks for the input!
This addresses #8 , that was previously assigned.
I personally think that this text is too much `tedana`, and we're actually using a different approach in general - that we can discuss together in an upcoming call.
Maybe it's better to rewrite it anew?

eurunuela:
I took it from `tedana` thinking most of it would be very similar. I removed what I thought would be not make sense in our case or whatever we have not implemented yet. I made the PR so that we all can discuss and suggest what the CONTRIBUTING.md should look like.

smoia:
Shall we close this one, following #84 ?

PR #76, opened 2019-12-04
[TST/WIP] Adds unit tests

eurunuela:
I'm creating this new draft PR as suggested by @smoia.

### DIFF
--- a/phys2bids/tests/test_phys2bids.py
+++ b/phys2bids/tests/test_phys2bids.py
@@ -1,0 +1,48 @@
+"""
+Tests phys2bids.py
+"""
+
+import os
+import json
+from phys2bids import phys2bids
+
+
+def test_print_summary(tmpdir):
+    test_filename = 'input.txt'
+    test_ntp_expected = 10
+    test_ntp_found = 5
+    test_samp_freq = 0.2
+    test_time_offset = 0.8
+    test_outfile = tmpdir.join('summary')
+    test_outfile_log = tmpdir.join('summary.log')
+    phys2bids.print_summary(test_filename, test_ntp_expected, test_ntp_found,
+                            test_samp_freq, test_time_offset, str(test_outfile))
+    assert os.path.isfile(str(test_outfile_log))
+    os.remove(str(test_outfile_log))
+
+
+def test_print_json(tmpdir):
+    test_outfile = tmpdir.join('json_test')
+    test_outfile_json = tmpdir.join('json_test.json')
+    test_samp_freq = 0.1
+    test_time_offset = -0.5
+    test_ch_name = 'foo'
+    phys2bids.print_json(str(test_outfile), test_samp_freq, test_time_offset, test_ch_name)
+    assert os.path.isfile(test_outfile_json)
+    test_json_data = dict(SamplingFrequency=0.1,
+                          StartTime=0.5,
+                          Columns='foo')
+    with open(test_outfile_json, 'r') as src:
+        loaded_data = json.load(src)
+    assert test_json_data == loaded_data
+    os.remove(str(test_outfile_json))
+
+
+def test_use_heuristic(tmpdir):
+    test_heur_file = 'heur_test_acq'
+    test_sub = 'SBJ01'
+    test_ses = 'S05'
+    test_filename = 'inputfile.txt'
+    test_outdir = tmpdir
+    test_record_label = 'test'
+    phys2bids.use_heuristic(str(test_heur_file), test_sub, test_ses, test_filename, test_outdir, test_record_label)
--- a/phys2bids/tests/test_phys2bids.py
+++ b/phys2bids/tests/test_phys2bids.py
@@ -62,5 +62,8 @@ def test_use_heuristic(tmpdir):
     heur_path = phys2bids.use_heuristic(test_full_heur_path, test_sub, test_ses,
                                         test_full_input_path, test_outdir, test_record_label)
 
-    assert f'_recording-{test_record_label}' in str(heur_path)
-    assert str(tmpdir) in str(heur_path)
+    test_result = ('/sub-SBJ01/ses-S05/func/sub-SBJ01_ses-S05_task-test_rec'
+                   '-biopac_run-00_recording-test_physio')
+    test_result = str(tmpdir) + test_result
+
+    assert test_result == str(heur_path)
--- a/phys2bids/tests/test_phys2bids.py
+++ b/phys2bids/tests/test_phys2bids.py
@@ -5,6 +5,7 @@ Tests phys2bids.py
 import os
 import json
 from phys2bids import phys2bids
+from pkg_resources import resource_filename
 
 
 def test_print_summary(tmpdir):
@@ -15,9 +16,12 @@ def test_print_summary(tmpdir):
     test_time_offset = 0.8
     test_outfile = tmpdir.join('summary')
     test_outfile_log = tmpdir.join('summary.log')
+
     phys2bids.print_summary(test_filename, test_ntp_expected, test_ntp_found,
                             test_samp_freq, test_time_offset, str(test_outfile))
+
     assert os.path.isfile(str(test_outfile_log))
+
     os.remove(str(test_outfile_log))
 
 
@@ -27,22 +31,36 @@ def test_print_json(tmpdir):
     test_samp_freq = 0.1
     test_time_offset = -0.5
     test_ch_name = 'foo'
+
     phys2bids.print_json(str(test_outfile), test_samp_freq, test_time_offset, test_ch_name)
+
     assert os.path.isfile(test_outfile_json)
+
     test_json_data = dict(SamplingFrequency=0.1,
                           StartTime=0.5,
                           Columns='foo')
     with open(test_outfile_json, 'r') as src:
         loaded_data = json.load(src)
+
     assert test_json_data == loaded_data
+
     os.remove(str(test_outfile_json))
 
 
 def test_use_heuristic(tmpdir):
-    test_heur_file = 'heur_test_acq'
+    test_heur_path = resource_filename('phys2bids', 'heuristics')
+    test_heur_file = 'heur_test_acq.py'
+    test_full_heur_path = os.path.join(test_heur_path, test_heur_file)
     test_sub = 'SBJ01'
     test_ses = 'S05'
-    test_filename = 'inputfile.txt'
+    test_input_path = resource_filename('phys2bids', 'tests/data')
+    test_input_file = 'Test_belt_pulse_samefreq.acq'
+    test_full_input_path = os.path.join(test_input_path, test_input_file)
     test_outdir = tmpdir
     test_record_label = 'test'
-    phys2bids.use_heuristic(str(test_heur_file), test_sub, test_ses, test_filename, test_outdir, test_record_label)
+
+    heur_path = phys2bids.use_heuristic(test_full_heur_path, test_sub, test_ses,
+                                        test_full_input_path, test_outdir, test_record_label)
+
+    assert f'_recording-{test_record_label}' in str(heur_path)
+    assert str(tmpdir) in str(heur_path)
--- a/phys2bids/tests/test_physio_obj.py
+++ b/phys2bids/tests/test_physio_obj.py
@@ -23,11 +23,60 @@ def test_has_size():
 def test_BlueprintInput():
     test_time = np.array([0, 1, 1, 2, 3, 5, 8, 13])
     test_trigger = np.array([0, 1, 0, 0, 0, 0, 0, 0])
-    test_timeseries = [test_time, test_trigger]
-    test_freq = [42.0, 3.14]
-    test_chn_name = ['Time', 'Trigger']
-    test_units = ['s', 's']
+    test_chocolate = np.array([1, 0, 0, 1, 0, 0, 1, 0])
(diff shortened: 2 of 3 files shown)


rmarkello:
So generally a draft PR means that it's *not* ready for review... Are you still working on this, or should I migrate all my pending comments from your old PR to this one?

eurunuela:
I drafted the PR as @smoia requested. I'm still working on it anyway. I've to write the unit tests for `phys2bids.py`. 

You'll have to migrate your comments from the old PR. All I've changed is a couple of incorrect paths.

smoia:
He's still working on them (hence the invitation to change PR #49 to a draft PR).

phys2bids/tests/test_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    test_path = '/home/travis/build/physiopy/phys2bids/phys2bids/tests/data'
+    assert utils.check_input_type(test_filename, test_path)
+
+
+# Tests path_exists_or_make_it
+def test_path_exists_or_make_it():
+    test_folder = '/home/travis/build/physiopy/phys2bids/tests/foo'
+    utils.path_exists_or_make_it(test_folder)
+    os.path.isdir(test_folder)
+    os.rmdir(test_folder)  # Removes the created folder
```
rmarkello: Can we add a test for a path that already exists? You should just be able to call the function again with the same input before you use `os.rmdir`!

Also, the way this is currently written precludes running the tests locally (for example, I won't be able to create the directory `'/home/travis/build` on my Mac laptop). To get around this, I would recommend using `pytest`'s [`tmpdir` fixture](http://doc.pytest.org/en/latest/tmpdir.html)!

phys2bids/tests/test_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    test_filename = 'help'
+    ext = 'py'
+    out_filename = utils.check_input_ext(test_filename, ext)
+    assert str(out_filename) == f'{test_filename}.{ext}'
+
+
+# Tests check_input_type
+def test_check_input_type():
+    test_filename = 'Test_belt_pulse_samefreq.acq'
+    test_path = '/home/travis/build/physiopy/phys2bids/phys2bids/tests/data'
```
rmarkello: We should probably be using `pkg_resource.resource_filename` to get this path so that we're not hard-coding this TravisCI path in our tests! Something like this should work and be a bit more robust to different build environments:

```python
from pkg_resources import resource_filename
test_path = resource_filename('phys2bids', 'tests/data')
```

phys2bids/tests/test_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    test_folder = '/home/travis/build/physiopy/phys2bids/tests/foo'
+    utils.path_exists_or_make_it(test_folder)
+    os.path.isdir(test_folder)
+    os.rmdir(test_folder)  # Removes the created folder
+
+
+# Tests check_file_exists
+def test_check_file_exists():
+    test_filename = 'Test_belt_pulse_samefreq.acq'
+    test_path = '/home/travis/build/physiopy/phys2bids/tests/data'
```
rmarkello: See my above comment on using `pkg_resources` for `test_check_input_type()`!

phys2bids/tests/test_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    test_full_path = os.path.join(test_path, test_filename)
+    utils.check_file_exists(test_full_path)
+
+
+# Tests move_file
+def test_move_file():
+    ext = '.txt'
+    test_path = '/home/travis/build/physiopy/phys2bids/tests/data'
```
rmarkello: Again, same deal with the `tmpdir` fixture from `test_path_exists_or_make_it()`! There are a few more instances of this here but I'm not going to comment on all of them.

phys2bids/tests/test_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    test_old_path = os.path.join(test_path, 'foo.txt')
+    open(test_old_path, 'a').close()
```
rmarkello:
```suggestion
    with open(test_old_path, 'a'):
        pass
```

might be a bit safer!

phys2bids/tests/test_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    os.remove(f'{test_new_path}.txt')
+
+
+# Tests copy_file
+def test_copy_file():
+    ext = '.txt'
+    test_path = '/home/travis/build/physiopy/phys2bids/tests/data'
+    test_old_path = os.path.join(test_path, 'foo.txt')
+    test_new_path = os.path.join(test_path, 'mrmeeseeks')
+    open(f'{test_old_path}.txt', 'a').close()
```
rmarkello:
```suggestion
    with open(f'{test_old_path}.txt', 'a'):
        passs
```

phys2bids/tests/test_utils.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+# Tests writejson
+def test_writejson():
+    test_path = '/home/travis/build/physiopy/phys2bids/tests/data'
+    test_json_filename = os.path.join(test_path, 'foo')
+    test_json_data = dict(SamplingFrequency=42,
+                          StartTime='00:00 ET',
+                          Columns='Rick')
+    utils.writejson(test_json_filename, test_json_data, indent=4, sort_keys=False)
```
rmarkello: Can you add a check after this function call to make sure the generated json file has been written to disk and "looks" correct? Maybe you could try reading it back in and make sure the round-trip went well! Something like (though perhaps not exactly):

```python
assert os.path.isfile(test_json_filename)
with open(test_json_filename, 'r') as src:
    loaded_data = json.load(src)
assert test_json_data == loaded_data
```

phys2bids/tests/test_physio_obj.py
```diff
@@ -0,0 +1,82 @@
...
+    test_list = [0, 1, 1, 2, 3, 5, 8, 13]
+    valid_output = po.is_valid(test_list, list, int)
+    assert valid_output == test_list
+
+# Tests has_size
+def test_has_size():
+    test_list = [0, 1, 1, 2, 3, 5, 8, 13]
+    test_length = 10
+    size_output = po.has_size(test_list, test_length, 0)
+    assert len(size_output) == test_length
```
rmarkello: Can we maybe add another check here to make sure that the data aren't padded if they're already the correct size? Also a few checks to make sure that we're padding the data with the correct fill value would be awesome!

phys2bids/tests/test_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    path_corrected = utils.check_input_dir(test_path)
+    assert path_corrected == test_path[:-1]
+
+
+# Tests check_input_ext
+def test_check_input_ext():
+    test_filename = 'help'
+    ext = 'py'
+    out_filename = utils.check_input_ext(test_filename, ext)
+    assert str(out_filename) == f'{test_filename}.{ext}'
```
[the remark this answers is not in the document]
eurunuela: Apparently, it doesn't. Good catch! I'vel opened #78.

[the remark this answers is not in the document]
smoia: Good catch guys. Thank you both!
Opened PR #77 and #78.

phys2bids/tests/test_utils.py
eurunuela: Thank you! I didn't know about this package.

phys2bids/tests/test_utils.py
eurunuela: I didn't know about this. Thank you for the heads-up @rmarkello !

eurunuela:
Any idea on how I could test `use_heuristics()` in `phys2bids.py`? @rmarkello @smoia

eurunuela:
Regarding my last question, do you think the latest commit is enough? @rmarkello @smoia

phys2bids/tests/test_phys2bids.py
```diff
@@ -0,0 +1,66 @@
...
+    test_input_path = resource_filename('phys2bids', 'tests/data')
+    test_input_file = 'Test_belt_pulse_samefreq.acq'
+    test_full_input_path = os.path.join(test_input_path, test_input_file)
+    test_outdir = tmpdir
+    test_record_label = 'test'
+
+    heur_path = phys2bids.use_heuristic(test_full_heur_path, test_sub, test_ses,
+                                        test_full_input_path, test_outdir, test_record_label)
+
+    assert f'_recording-{test_record_label}' in str(heur_path)
```
rmarkello: Shouldn't we be able to determine exactly what the generated filename will be? It might be nice to test against that in this case rather than just checking a sub-string.

phys2bids/tests/test_physio_obj.py
```diff
@@ -0,0 +1,123 @@
...
+    assert test_index[1] == len(test_timeseries)
+    assert test_index[3] == test_chn_name[1]
+    assert test_index[4] == test_units[1]
+
+    # Tests delete_at_index
+    blueprint_out.delete_at_index(1)
+    assert len(blueprint_out.ch_name) == num_channnels - 1
+    assert len(blueprint_out.units) == num_channnels - 1
+    assert blueprint_out.timeseries.shape[0] == num_channnels - 1
+    assert blueprint_out.ch_amount == num_channnels - 1
```
rmarkello: This is currently failing the test because `BlueprintOutput` is not updating `ch_amount` in the `.delete_at_index()` method. I'm fine to merge this PR in and then issue a bug fix afterwards, but it's worth noting!

eurunuela: I already opened an issue on this. Check #88

phys2bids/tests/test_phys2bids.py
eurunuela: You're right. I'll make a quick commit.

phys2bids/tests/test_phys2bids.py
```diff
@@ -0,0 +1,69 @@
...
+    test_full_input_path = os.path.join(test_input_path, test_input_file)
+    test_outdir = tmpdir
+    test_record_label = 'test'
+
+    heur_path = phys2bids.use_heuristic(test_full_heur_path, test_sub, test_ses,
+                                        test_full_input_path, test_outdir, test_record_label)
+
+    test_result = ('/sub-SBJ01/ses-S05/func/sub-SBJ01_ses-S05_task-test_rec'
+                   '-biopac_run-00_recording-test_physio')
+    test_result = str(tmpdir) + test_result
```
rmarkello: This _may_ cause problems for Windows users, but we can tackle all those possible Windows issues when we add that testing. Thanks for updating this!

rmarkello: Also: I think `run-XX` is supposed to be 1-indexed... Perhaps we should fix that in `use_heuristic`. (I'll check the BIDS spec.)

rmarkello:
Shall I go ahead and merge this in, or do you want to wait for @smoia's review?

phys2bids/tests/test_phys2bids.py
smoia: I didn't see the one indexing as a must, however the examples are one indexed. If you find it in the docs, let me know, otherwise let' not force it.

smoia: Ah, wait, it's [here](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#participant-names-and-other-labels).
```
BIDS uses custom user-defined labels in several situations (naming of participants, sessions, acquisition schemes, etc.) Labels are strings and MUST only consist of letters (lower or upper case) and/or numbers. If numbers are used we RECOMMEND zero padding (e.g., 01 instead of 1 if you have more than nine subjects) to make alphabetical sorting more intuitive.

Please note that a given label is distinct from the "prefix" it refers to. For example sub-01 refers to the sub entity (a subject) with the label 01. The sub- prefix is not part of the subject label, but must be included in file names (similarly to other key names). In contrast to other labels, run and echo labels MUST be integers. Those labels MAY include zero padding, but this is NOT RECOMMENDED to maintain their uniqueness.
```

I'd go for a "let the user decide", but if you want to add a check for that, I don't see any problem either. @rmarkello , opinions?

eurunuela:
I wanted to wait on @smoia's input. He might have some feedback for us.

phys2bids/tests/test_phys2bids.py
rmarkello: This falls a bit on the side of personal preference&mdash;in my mind using a 1-index for `run` makes a bit more sense than 0-index, but I'll take a look at how it's specified in `use_heuristic` and we can see if it's worth modifying. I'll open an issue so we can continue discussion outside the PR.

smoia:
For the sake of timing, you can go on with the merge: I trust the two of you more than me.
Quick point: @rmarkello , is it practice to have docstrings for the unit tests or not? Should we enforce it?
If not, up to one of you to do the honours!

eurunuela:
I'll merge it then. We can always add the docstrings in a documentation related PR. I think it's important we have the test running on every PR we do from now on.

rmarkello:
Sorry, somehow missed this ping @smoia: no, doc-strings are generally not expected for tests. I believe the thought is that it's clear from the test name what function is being tests; however, liberal use of comments throughout the body of the test function are definitely useful! We can add those in as the test functions are built on.

:raised_hands: for merging this in, though ! Thanks @eurunuela !

PR #79, opened 2019-12-06
Bugfixes

### LINKED ISSUE #77: [BUG] Cannot delete array elements in Blueprint_Output.delete_at_index()
Line raising the error: https://github.com/physiopy/phys2bids/blob/46f5239d7ef9f4722394f1a0c9c4a28929f4c0b2/phys2bids/physio_obj.py#L397

Error:

```
blueprint_out.delete_at_index(1)...
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/phys2bids/phys2bids/tests/test_physio_obj.py in 
----> 1 blueprint_out.delete_at_index(1)
      2 assert blueprint_out.ch_amount == num_channnels - 1

~/phys2bids/phys2bids/physio_obj.py in delete_at_index(self, idx)
    395             `idx` gets deleted
    396         """
--> 397         del(self.timeseries[idx])
    398         del(self.ch_name[idx])
    399         del(self.units[idx])

ValueError: cannot delete array elements
```

### LINKED ISSUE #78: [BUG] check_input_ext doesn't work with double extensions
As the title says, the method does not work with double extensions when the extension given is not `.gz`. Have a look at the example below:

```
check_input_ext('help.txt.gz', 'py') != 'help.py'
```

smoia:
Fixes #77 and fixes #78 

Corrected method `delete_at_index` in `BlueprintOutput` to deal with numpy arrays
Corrected function in `utils.py` `check_input_ext`, now it can correctly handle `.gz` and `.ext.gz` extensions.

Is there any other possible double extension we should be able to check?

### DIFF
--- a/phys2bids/utils.py
+++ b/phys2bids/utils.py
@@ -23,15 +23,13 @@ def check_input_ext(filename, ext):
     """
     Checks that the given file has the given extension
     """
-    if '.gz' in ext:
-        if filename[-len(ext):] != ext:
-            filename = filename + ext
-        return filename
-    else:
-        if ext[0] != '.':
-            ext = '.' + ext
+    if '.gz' in filename:
+        filename = filename[:-3]
 
-        return Path(filename).with_suffix(ext)
+    if ext[0] != '.':
+        ext = '.' + ext
+
+    return Path(filename).with_suffix(ext)
 
 
 def check_input_type(filename, indir):
--- a/phys2bids/utils.py
+++ b/phys2bids/utils.py
@@ -13,7 +13,7 @@ def check_input_dir(indir):
     """
     Checks that the given indir doesn't have a trailing '/'
     """
-    if indir[-1:] == '/':
+    if indir.endswith('/'):
         indir = indir[:-1]
 
     return indir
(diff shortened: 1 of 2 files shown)


smoia:
As addressed in #76

phys2bids/utils.py
```diff
@@ -23,15 +23,13 @@ def check_input_ext(filename, ext):
...
     """
-    if '.gz' in ext:
-        if filename[-len(ext):] != ext:
-            filename = filename + ext
-        return filename
-    else:
-        if ext[0] != '.':
-            ext = '.' + ext
+    if '.gz' in filename:
+        filename = filename[:-3]
```
rmarkello: What if we maybe do:

```python
if filename.endswith('.gz'):
    filename = filename[:-3]
```

This might be a bit safer in the (unlikely) event that someone has a file like `file.py.gz.txt` (or something else equally weird).

phys2bids/utils.py
```diff
@@ -23,15 +23,13 @@ def check_input_ext(filename, ext):
...
+
+    if ext[0] != '.':
+        ext = '.' + ext
```
rmarkello: This is slightly preference, but I think using the string methods for these checks might be a bit cleaner:

```python
if not ext.startswith('.'):
    ext = '.' + ext
```

Issue #83, opened 2019-12-07
default channel trigger doesn't work for txt files


vinferrer:
Hi, doing some testing , i have realized that the default of  -chtrig is not vaild for labchart txt files since the trigger channel happens to be index 1 no index 0, should we change default or only correct it when txt files are used?

smoia:
We don't know yet if all the txt files have the trigger in position 1. Actually, we might expect it to be all over the place.

A better approach would be to force the user to put it, *if* the option `-info` is not present.
For instance, we can set its default to `None`, and raise an exception if `info` is not True and `chtrig` is None.

Another approach would be to *not* consider the first channel when processing labchart files - so that we can propose to standardise the trigger as first channel during acquisition, and not break the code.

What do you think @RayStick @eurunuela @rmarkello  (and anyone else with an opinion in the matter, ofc)?

RayStick:
Yes, I noticed this as a problem too. In our labchart txt files the first column [0] is time, and the second column is the trigger [1]. In these txt files, the first column will always be time, but the second column won't always be trigger. It normally is, but if depends completely on how the user has set up their hardware.

RayStick:
As to the solution, I think either of the things you suggest would work @smoia. Forcing the user to input it may be best, at least for now.

RayStick:
I've noticed that we have some Chicago data labelled 'multiscan' and some labelled ' 2minrest' so best to test on the '2minrest' for now. 

I have also put a bit more Chicago data up on OSF (labelled ‘Test2’) in case it helps with this testing. One scan exported from lab chart in 4 different ways: header with time column, header with no time column, no header with time column, no header with no time column. I have included the number of TRs in the file name, so you know if phys2bids is getting it right.

vinferrer:
Thank you @RayStick , from now on we can adress all things related to this issue in draft PR #85,  there you will see a photo with the newest output of the plot function.

smoia:
Moved this converation to #118

PR #85, opened 2019-12-07
[ENH] Adding subplots and correcting tiny issue with txt 

vinferrer:


vinferrer:
Hi guys, here is the code for subplots, i also added a tiny change for the subplot indexes, I leave you how the plot looks right now:
image

vinferrer:
So @smoia, @RayStick , right now this is implemented once the ``phys_in`` object is already populated. Do we want it this way or do we want the function to deal directly with the input? that can quite challenging since each format can present the information in different manners which would mean basically we would be force to implement a plotting function for each format.  On the other hand if the user doesn't know which are the inputs order he will be able to guess better this way

vinferrer:
Another idea would be using the function as it is now, but adding the original channel order as information. This way we would have only one function that can deal with our standart ``phys_in`` and at the same time the user nows the original order of the channels

smoia:
That's a possible solution! The only channel we move around is the trigger, and we know where it is due to user input - so that can be achieved.
Another solution is that when we populate the BlueprintInput object, we don't care where the trigger channel is, but then we write a class function that moves it in front. In this way we also remove "chtrig" from the interfaces scripts. Having the trigger as first column is not even necessary, but it helps standardising the data a bit for later uses.

vinferrer:
So, what could we go for?

smoia:
The first one is easier - I guess.
The second is probably more correct?
@eurunuela @rmarkello what do you think?

RayStick:
My [user testing](https://docs.google.com/document/d/1XPKg7gBFocbiy58KFT6PWjdQjdxnxFSZL_TbnsQT8g4/edit#) may be relevant here (see last entry)
I can't get phys2bids to work with labchart txt files, but I gather this is what is being addressed in this PR, due to the bug in #83.

eurunuela:
I'd avoid creating a plotting function per input format and provide the original channel order, following @vinferrer 's second suggestion.

vinferrer:
If we go for my option, we still have to modify the input reading functions so they gather the original order of the channels since this info is not in ``phys_in``

vinferrer:
Change the fig size to the default, maximum of seconds (half hour), interpolate to a certain frequency

vinferrer:
Hey guys, so I changed the file I use to plot, I am using now the 2minRest, apart from that I have put as default FIGSIZE for the size of the figure, limit the recording plot to 30 min and unified the x axis:

image

vinferrer:
This is nice because for labchart files we don't need to deal with the multiple frequencies problem

vinferrer:
@smoia do you have the ``acq`` file with multiple frequency?

smoia:
This looks good to me! Let me check where did I save it.

smoia:
[here you go!](https://mfr.de-1.osf.io/render?url=https://osf.io/9a7yv/?action=download%26mode=render)
It's in OSF. It's laso exactly the same as the other one in OSF in the same folder, except for one channel that has different frequency sample.

phys2bids/viz.py
```diff
@@ -41,3 +41,21 @@ def ntr2time(x):
     subplot.plot(time, trigger, '-', time, time, '-')
     plt.savefig(outfile + '_trigger_time.png', dpi=dpi)
     plt.close()
+
+def plot_all(phys_in,infile,outfile,dpi = SET_DPI,size = FIGSIZE):
+    ch_num = len(phys_in.ch_name) # get number of channels:
+    fig,ax = plt.subplots(ch_num-1,1,figsize = size,sharex=True)
+    row = 0
+    time = phys_in.timeseries[0] # assume time is first channel
+    fig.suptitle(infile)
```
smoia: Why don't we return just the filename?
Also, it would be great if we can make this a big bigger (2 times the size it is now).
```suggestion
    fig.suptitle(os.path.basename(infile))
```

phys2bids/viz.py
```diff
@@ -41,3 +41,21 @@ def ntr2time(x):
     subplot.plot(time, trigger, '-', time, time, '-')
     plt.savefig(outfile + '_trigger_time.png', dpi=dpi)
     plt.close()
+
+def plot_all(phys_in,infile,outfile,dpi = SET_DPI,size = FIGSIZE):
+    ch_num = len(phys_in.ch_name) # get number of channels:
+    fig,ax = plt.subplots(ch_num-1,1,figsize = size,sharex=True)
```
smoia: This is not going to pass the lintering test

phys2bids/viz.py
```diff
@@ -41,3 +41,21 @@ def ntr2time(x):
...
+    row = 0
```
smoia: I saw this, but is it really necessary? You can just use `index` in the loop.

phys2bids/viz.py
```diff
@@ -41,3 +41,21 @@ def ntr2time(x):
...
+    time = phys_in.timeseries[0] # assume time is first channel
+    fig.suptitle(infile)
+    for index, timeser in enumerate(phys_in.timeseries[1:]):
+        index += 1
```
smoia: This too might not be necessary: why don't you use `index` instead of dealing with `row` (if you think it's more clear, use the name`row`), and correct indexing where you need to?

RayStick:
@vinferrer new figure looks great.

phys2bids/interfaces/txt.py
```diff
@@ -67,7 +66,7 @@ def populate_phys_input(filename, chtrig):
     # get names
     orig_names=header[4][1:]
     names = ['time','trigger']
-    orig_names.pop(chtrig)
+    orig_names.pop(chtrig-1)
```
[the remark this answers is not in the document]
vinferrer: Because orignames starts at 1 instead of 0 and therefore the first channel which is time is not taken into account

vinferrer:
> [here you go!](https://mfr.de-1.osf.io/render?url=https://osf.io/9a7yv/?action=download%26mode=render)
> It's in OSF. It's laso exactly the same as the other one in OSF in the same folder, except for one channel that has different frequency sample.

This link is useless i can't see the file neither dowload it

smoia:
It's still in phys2bids OSF, but I'll change the link to be useful.

vinferrer:
Just showing how the acq file looks, we should enforce -chtrig as needed argument for now:
image

vinferrer:
Good news, multifrequency support is added to the plotting, however we need to put as requirements to install ``scipy``  since i use one function from there for the resampling

phys2bids/phys2bids.py
```diff
@@ -212,8 +212,10 @@ def _main(argv=None):
     print('Reading infos')
     phys_in.print_info(options.filename)
     # #!# Here the function viz.plot_channel should be called
-    # for the desired channels.
-
+    if options.chplot[0] !=  "" :
```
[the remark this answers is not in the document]
vinferrer: make change because now it is a string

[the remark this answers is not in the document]
vinferrer: Sry the thing is, that if you input a path this variable becomes a list, and therefore the default must also be a list so everything checks out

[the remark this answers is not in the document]
vinferrer: As for the not making it optional, we can do that but in my opinion, i user i don't appreciate that things generate files without asking for them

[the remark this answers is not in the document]
vinferrer: My opinion is that the user should ask for the plot and give a name or dir to the file as it is now

vinferrer:
The multifrequency support has only been tried with one acq file where the sampling is inferior to time, this should be check with other files where sampling is bigger just in case

smoia:
> Good news, multifrequency support is added to the plotting, however we need to put as requirements to install `scipy` since i use one function from there for the resampling

That is great, but would it be possible to use `numpy.interp` instead? It's true that `numpy` and `scipy` are frequently installed together and probably are there already, but we refactored the code to remove dependencies, and I think it's a good way to go.

vinferrer:
> > Good news, multifrequency support is added to the plotting, however we need to put as requirements to install `scipy` since i use one function from there for the resampling
> 
> That is great, but would it be possible to use `numpy.interp` instead? It's true that `numpy` and `scipy` are frequently installed together and probably are there already, but we refactored the code to remove dependencies, and I think it's a good way to go.

The numpy.interp function is not what we are looking for, plus this ``resample`` can upsample and downsample, therefore in one if you do both things, i think using ``np.interp`` complicates the work

rmarkello:
I agree with @smoia here, and in fact would go a bit further and argue that `scipy.signal.resample` is the _wrong_ method for this because it's using a Fourier-based resampling method when what we really probably want is a linear interpolation. Looking at the example in the [`scipy.signal.resample` docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample.html) it's clear that the resampling is doing weird things towards the lower x-values (notice it introduces fluctuations that aren't present in the original data). On the other hand, `numpy.interp` (because it's a linear interpolation) _can't_ introduce such artifacts.

Also, `numpy.interp` can downsample (in addition to upsample) data:

```python
>>> import numpy as np
>>> x = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> y = [3, 2, 1, 2, 3, 2, 1, 3, 2, 1]
>>> np.interp([0, 0.5, 1], x, y)
array([3. , 2.5, 2. ])
```

Since this is purely for visualization purposes I'm worried that the arbitrary fluctuations that can be induced by `scipy.signal.resample` might be confusing (and, as @smoia notes, it's an extra dependency). 

Nonetheless, I'm happy to be convinced otherwise here, though! Let me know your thoughts

rmarkello:
(I don't want the point about `scipy` to detract from the fact that this PR is *awesome* and the plots are looking great!!!!! :raised_hands: :tada: Thanks so much for all the hard work, @vinferrer!)

vinferrer:
@smoia this is ready for merge please review whatever you want but accorded changes are done

phys2bids/viz.py
```diff
@@ -20,7 +21,7 @@ def time2ntr(x):
         return x / options.tr
 
     def ntr2time(x):
-        return x * options.tr
+        return x * options.trprint(f'saving channels plot at plot at {options.chplot}')
```
smoia: I think there's a mistake here?
the print should go two lines later one indentation level less!

README.md
```diff
@@ -16,6 +16,7 @@ It doesn't support physiological files recorded with the MRI, as you can find a
 It requires python 3.6 or above, as well as the modules:
 - `numpy >= 1.9.3`
 - `matplotlib >= 3.1.1`
+- `scipy==1.3.3`
```
smoia: Can you address @rmarkello 's answer in the PR?

phys2bids/phys2bids.py
```diff
@@ -212,8 +212,9 @@ def _main(argv=None):
     print('Reading infos')
     phys_in.print_info(options.filename)
     # #!# Here the function viz.plot_channel should be called
-    # for the desired channels.
-
+    if options.chplot != '' or options.info:
+        from phys2bids.viz import plot_all
```
smoia: We're already importing viz at line 37!
```suggestion

```

phys2bids/phys2bids.py
```diff
@@ -212,8 +212,9 @@ def _main(argv=None):
...
+        plot_all(phys_in, infile, options.chplot)
```
smoia:
```suggestion
        viz.plot_all(phys_in, infile, options.chplot)
```

phys2bids/phys2bids.py
```diff
@@ -212,8 +212,9 @@ def _main(argv=None):
     print('Reading infos')
     phys_in.print_info(options.filename)
     # #!# Here the function viz.plot_channel should be called
-    # for the desired channels.
-
+    if options.chplot != '' or options.info:
```
smoia: If `option.info`, but `options.chplot == ''`, there won't be a path to save the plot.
```suggestion
    if options.chplot != '' or options.info:
        if options.chplot == '':
            options.chplot == f'{infile}.png'
```

smoia:
@vinferrer can you address @rmarkello 's point about `numpy.interp`?

@rmarkello , could you review this once @vinferrer did the scipy to numpy change?

Thank you guys!

> I agree with @smoia here, and in fact would go a bit further and argue that `scipy.signal.resample` is the _wrong_ method for this because it's using a Fourier-based resampling method when what we really probably want is a linear interpolation. Looking at the example in the [`scipy.signal.resample` docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample.html) it's clear that the resampling is doing weird things towards the lower x-values (notice it introduces fluctuations that aren't present in the original data). On the other hand, `numpy.interp` (because it's a linear interpolation) _can't_ introduce such artifacts.
> 
> Also, `numpy.interp` can downsample (in addition to upsample) data:
> 
> ```python
> >>> import numpy as np
> >>> x = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
> >>> y = [3, 2, 1, 2, 3, 2, 1, 3, 2, 1]
> >>> np.interp([0, 0.5, 1], x, y)
> array([3. , 2.5, 2. ])
> ```
> 
> Since this is purely for visualization purposes I'm worried that the arbitrary fluctuations that can be induced by `scipy.signal.resample` might be confusing (and, as @smoia notes, it's an extra dependency).
> 
> Nonetheless, I'm happy to be convinced otherwise here, though! Let me know your thoughts

phys2bids/phys2bids.py
vinferrer: This is already managed inside the function don't worry about it

phys2bids/viz.py
vinferrer: actually i don't know how that happened but is taken care of

phys2bids/viz.py
```diff
@@ -41,3 +41,25 @@ def ntr2time(x):
...
+
+def plot_all(phys_in, infile, outfile='', dpi=SET_DPI, size=FIGSIZE):
+    ch_num = len(phys_in.ch_name)  # get number of channels:
+    fig, ax = plt.subplots(ch_num - 1, 1, figsize=size, sharex=True)
+    time = phys_in.timeseries[0]  # assume time is first channel
+    fig.suptitle(os.path.basename(infile))
+    for row, timeser in enumerate(phys_in.timeseries[1:]):
+        if timeser.shape != time.shape:
+            time_old = np.linspace(0, time[-1], num=timeser.shape[0])
+            timeser = np.interp(time, time_old, timeser)
```
vinferrer: Well guys here you have it, interpolation with np.interp. I already delete any presence of scipy so this should be now ready for merging

eurunuela: Great! Thanks @vinferrer !

phys2bids/viz.py
```diff
@@ -41,3 +41,25 @@ def ntr2time(x):
...
+        ax[row].plot(time, timeser)
+        ax[row].set_title(f' Channel {row + 2}: {phys_in.ch_name[row + 1]}')
```
smoia:
```suggestion
        ax[row].set_title(f' Channel {row + 1}: {phys_in.ch_name[row + 1]}')
```

phys2bids/viz.py
```diff
@@ -41,3 +41,25 @@ def ntr2time(x):
...
+        ax[row].set_ylabel(phys_in.units[row + 1])
+        ax[row].xlim = 30 * 60 * phys_in.freq[0]  # maximum display of half an hour
+        ax[row].grid()
+        row += 1
```
smoia:
```suggestion

```

phys2bids/viz.py
```diff
@@ -41,3 +41,25 @@ def ntr2time(x):
...
+    ax[row - 1].set_xlabel("seconds")
```
smoia:
```suggestion
    ax[row].set_xlabel("seconds")
```

PR #93, opened 2019-12-11
Proposal of issue template

eurunuela:
I have based it on tedana's issue template as I believe it aligns with what we want here.

### DIFF
--- a/.github/ISSUE_TEMPLATE/custom.md
+++ b/.github/ISSUE_TEMPLATE/custom.md
@@ -1,0 +1,42 @@
+---
+name: Custom issue template
+about: Possible issue template for phys2bids
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+<!--
+This is a suggested issue template for phys2bids.
+
+If there is other information that would be helpful to include, please do not hesitate to add it!
+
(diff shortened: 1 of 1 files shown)


smoia:
I would check out the standard templates of GitHub (we can even have one for bugs, one for others, ...) too! Also, you can check out templates from other projects here in GH. For instance, all-contributors, numpy, nibabel, heudiconv, ...

If you do copy a template from any source (and it's not standard GitHub template, please report attribution with the correct copyright and link). Pay attention to the fact that we might need to adopt a different licence than Apache2.0, according to your source licence - that would be unconvenient, especially if it's for a structural thing.

eurunuela:
I did check templates of other projects on GitHub and they’ve all very similar. They do not use one template for bugs, another one for feature requests, etc. 

We could have one per type of issue if you want, but I don’t think it’s worth it. I’d have our custom one plus maybe one for bugs.

smoia:
Cool, as you prefer!
Let's check the thing of the licence though - tedana is LGPL 2.1.

eurunuela:
We’re talking about an issue template. Is the license such a big deal for this?

eurunuela:
I have found these free of any license: https://github.com/stevemao/github-issue-templates

Issue #74, opened 2019-12-03
README - instructions for package install


RayStick:
It may be obvious to regular python users, but it may help to instruct the user to clone the repo into the location where their python packages are located. For me, I cloned it elsewhere first, and then when I later renamed a directory the path to the package was wrong. So just some extra text in the REAME suggesting the location of where this should go (i.e. for me on a macOS it was **/usr/local/lib/python3.7/site-packages** ). If there is a reason not to put it in site-packages, let me know.

vinferrer:
Hi, there is no reason to put it in the location where the site-packages are, actually this is not recommended at all, like last thing you should do. The dir is irrelevant since is the setup.py scritp the one that manages to put the libraries in the right place when installing, but the repo should be in the home dir, for example

RayStick:
I did wonder whether it was a good idea or not. Can you elaborate on why it is the last thing I should do? I understand I would want to clone the repo to my home dir, for example, if it is development code. However when I run the suggested line 'sudo pip3 install -e' in the README.md file that's fine as long as I do not change _any_ of the paths to this directory. If I do, the phys2bids package does not work. How can I avoid that?

What does the '-e' mean in the **sudo pip3 install -e .** command mean? 
Also, I assumed '.' meant current directory, so I tried to change that to a different location but got an error

eurunuela:
I believe that `-e` is just the _key_ that makes pip understand that you're giving a path as an argument (see https://pip.pypa.io/en/stable/reference/pip_install/).

What do you mean by `pip3 install` works as long as you do not change _any_ of the paths to _this_ directory? I don't think I understand the issue here.

RayStick:
Ah, I understand -e now. 

In terms of location, let me see if I can rephrase (please bear in mind I am relatively new to using python): 

I guess I am just trying to work out the 'best' way to interact with this package/have it set up on my computer. I am most familiar with installing packages with pip3 which appears to put them into my default location of **/usr/local/lib/python3.7/site-packages**. I am wondering why this phys2bids package shouldn't go into this location as well?

eurunuela:
So, pip will usually install a package from a url and into a specific Python packages folder on your computer. 

In our case, we're telling pip to install the package located in the given folder (i.e. the phys2bids project) as a new "system command". 

[...]

PR #94, opened 2019-12-11
[DOC] Update README.md

rmarkello:
Relates to #74.

Removes recommendation for using `sudo` and `-e` during package installation. I'm open to discussion on the option of the `-e` flag, but feel very strongly about **not** recommending the `sudo`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Donwload the package as zip from github and uncompress or if you have ``git`` us
 
 open a terminal in the phy2bids folder and execute the command:
 
-``sudo pip3 install -e .``
+``pip3 install .``
 
 type the command:
 


eurunuela:
I'd add the `sudo` and `-e` to CONTRIBUTING.md following the discussions on #74.

smoia:
While I agree with you @rmarkello , @vinferrer had a very strong opinion about *having* the `sudo` there.
I think he and @eurunuela tested the isntallation with and without `sudo` and it worked only with.
May I suggest a middle ground in which we:
1. Talk about the developer installation
2. Talk about the user installation (that might even be *without* pip)
3. On both sides, we describe them without sudo but say it might be needed?

eurunuela:
@vinferrer and I could not make it work without `sudo` that's right. 

As I was saying, we should definitely differentiate between the developer and user installation. The developer one should probably go into CONTRIBUTING.md while the user one should go into README.md? What do you think @rmarkello  @smoia ?

rmarkello:
Requiring `sudo` for `pip` suggests that either (1) something is a bit strange about the Python installation (i.e., you are trying to install to your system-wide site-packages, which are protected&mdash;for a good reason!&mdash;and thus `sudo` is required) or (2) we have set up the installation incorrectly somehow. I'd err towards (1) since the installation is working on TravisCI despite not having `sudo` in the build instructions.

An alternative to the `sudo` install that is much safer and should work regardless of your Python setup is to use `pip install --user .`. Could @eurunuela and @vinferrer try that and ensure it works on their systems (without `sudo`)? I would be happy to suggest this as an alternative for users and developers alike (developers can use the `--editable` flag in addition to the `--user` flag!).

rmarkello:
(Note, you'll likely have to uninstall your current version of `phys2bids` before trying the `pip install --user --editable .` to check that it works. If you installed with `sudo` you may need to use `sudo` to uninstall, as well.)

eurunuela:
I confirm `pip install .` worked on my Mac. @vinferrer could you try on your linux?

vinferrer:
so this is weird, i cannot unistall ``phys2bids``, ``pip3`` says it doesn't exists, and when i do ``pip3 freeze`` there is a library called: ``-e git+git@github.com:vinferrer/phys2bids.git@4aee1632590c545a8285028f16fd430df725b3e4#egg=phys2bids`` what do I do? Do I have to delete the git directory? @rmarkello

vinferrer:
Hey, so i managed to uninstall phys2bids, however when I try to install again without ``sudo`` and this happens:

```
pip3 install --user --editable .

Obtaining file:///home/vicente/phys2bids
    Complete output from command python setup.py egg_info:
    running egg_info
    /home/vicente/.local/lib/python3.6/site-packages/setuptools/dist.py:475: UserWarning: Normalizing 'v1.1.0-beta+103.g4aee163' to '1.1.0b0+103.g4aee163'
      normalized_version,
    error: [Errno 13] Permission denied
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /home/vicente/phys2bids/
```
if i install with sudo everything is fine. Maybe is a problem with my python configuration as @rmarkello  says

eurunuela:
@vinferrer is there any way you could check if it's a problem with your installation?

vinferrer:
I mean i don't know if it is a only my problem, @smoia has a linux, he can try to install it without ``sudo``

RayStick:
Sorry for my slow reply 
`pip install .` worked on my MacOS
(I had to uninstall phys2bids with sudo as well)

rmarkello:
Sorry for the delayed response here! I was travelling Friday through Mon and am only just catching up from being away for four days :see_no_evil: 

@vinferrer: I think this may have been caused by how you installed Python, but it's tough to say for sure. If you'd be willing, could you try running `sudo chown -R vicente ~/.local && chmod -R 755 $(find ~/.local -type d)` and then do the `pip install --user --editable .` from inside the `phys2bids` directory? I think for some reason your `~/.local` (or, more specifically, `~/.local/lib/python3.6`) directory permissions are wrong (possibly caused by use of `sudo` when you were installing Python or Python packages?), and this *should* fix that! :crossed_fingers: Let me know how it goes or if you run into other problems.

@eurunuela and @RayStick : I'm glad it worked for you! :+1: :tada:

smoia:
Can we delete this branch?

rmarkello:
@smoia: done!

@vinferrer: were you able to resolve the issues with installing?

smoia:
@rmarkello thank you!

PR #100, opened 2019-12-16
[FIX] Fixes linting errors

### LINKED ISSUE #92: Correcting linter errors
We have a bunch of errors and warnings in the linter test that makes it fail.
While I don't see any problem for Warnings (but maybe I'm wrong), we can't have Errors.
looking at Travis CI details, I see most of them are in `interfaces/txt.py`. @vinferrer , can you check that your scripts pass lintering for Travis CI?

eurunuela:
Partially fixes #92

## Proposed Changes

  - Fixes linting errors in `physio_obj.py`
  - Fixes linting errors in `cli/run.py`

### DIFF
--- a/phys2bids/cli/run.py
+++ b/phys2bids/cli/run.py
@@ -41,20 +41,20 @@ def _get_parser():
                           dest='outdir',
                           type=str,
                           help='Folder where output should be placed. '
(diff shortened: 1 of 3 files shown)


eurunuela:
#85 should fix the remaining linting errors.

PR #126, opened 2020-01-10
[DOC] populated doc with outline and initial info

AyyagariA:
Closes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/docs/bestpractice.rst
+++ b/docs/bestpractice.rst
@@ -4,7 +4,7 @@
 Best Practices for Collecting Physiological Data
 ================================================
 
-- This document is a work in progress. Currently, an outline of important topics is presented.
+This document is a work in progress. Currently, an outline of important topics is presented.
 
 .. _whycollectphysdata:
 
--- a/docs/bestpractice.rst
+++ b/docs/bestpractice.rst
@@ -37,7 +37,7 @@ How to collect each type of physiological data?
 
 .. _whattodowithphysdata:
 
-*What to do with physiological data once it has been collected?*
+What to do with physiological data once it has been collected?
 --------------------------------------------------------------------
 
 - Use phys2bids to organize the data
--- a/docs/bestpractice.rst
+++ b/docs/bestpractice.rst
@@ -8,7 +8,7 @@ Best Practices for Collecting Physiological Data
 
 .. _whycollectphysdata:
 
-*Why collect physiological data?*
+Why collect physiological data?
 ----------------------------------
 - Physiological fluctuations can be treated as noise in fMRI data or as signals of interest to characterize different physiological states
 - Add regressors in the modeling of fMRI data to explain signal variance
--- a/docs/bestpractice.rst
+++ b/docs/bestpractice.rst
@@ -19,7 +19,7 @@ Why collect physiological data?
 
 .. _differenttypesofphysdata:
 
-*What are the different types of physiological data that can be collected?*
+What are the different types of physiological data that can be collected?
 ----------------------------------------------------------------------------
 
 - CO2& O2: specifically end-tidal pressures of CO2 and O2, which approximate the arterial levels of these gases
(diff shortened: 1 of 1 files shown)


docs/bestpractice.rst
```diff
@@ -4,4 +4,33 @@
 Best Practices for Collecting Physiological Data
 ================================================
 
-Coming soon.
```
smoia: Could you add a quick text at the beginning that says something in the lines "we're working on it, but here's the content we're going to write down" or similar?

docs/bestpractice.rst
```diff
@@ -4,4 +4,33 @@
...
\ No newline at end of file
+**Why collect physiological data?**
```
smoia: It's not clear if you want just bold, or you want to create subtitles - in the second case, can you adapt them following [this document](https://github.com/ralsina/rst-cheatsheet/blob/master/rst-cheatsheet.rst#section-structure)? It would make it more clear even in the source code!
A cool thing that you can also add is an anchor before every title, so that we can later link to each specific part! Look, for instance, [here](https://stackoverflow.com/a/19543591).

AyyagariA: I intentionally made the text bold. Would you like it to be italic headings with only the document title in bold? I can change it to that if you would like - let me know!

docs/bestpractice.rst
AyyagariA: Yup!

docs/bestpractice.rst
```diff
@@ -4,4 +4,46 @@
 Best Practices for Collecting Physiological Data
 ================================================
 
-Coming soon.
\ No newline at end of file
+- This document is a work in progress. Currently, an outline of important topics is presented.
```
smoia:
```suggestion
This document is a work in progress. Currently, an outline of important topics is presented.
```

docs/bestpractice.rst
```diff
@@ -4,4 +4,46 @@
...
+
+.. _whycollectphysdata:
+
+*Why collect physiological data?*
```
smoia:
```suggestion
Why collect physiological data?
```

docs/bestpractice.rst
```diff
@@ -4,4 +4,46 @@
...
+- Physiological fluctuations can be treated as noise in fMRI data or as signals of interest to characterize different physiological states
+- Add regressors in the modeling of fMRI data to explain signal variance
+
+  - Especially cardiac and respiratory effects as they can influence in the BOLD signal
+- Cardiac pulsatility and breathing related chest/abdomen movement --> RETROICOR
+- Monitor subject/patient during the scan (important during first-time scans, or breathing tasks)
+
+.. _differenttypesofphysdata:
+
+*What are the different types of physiological data that can be collected?*
```
smoia:
```suggestion
What are the different types of physiological data that can be collected?
```

docs/bestpractice.rst
```diff
@@ -4,4 +4,46 @@
...
+----------------------------------------------------------------------------
+
+- CO2& O2: specifically end-tidal pressures of CO2 and O2, which approximate the arterial levels of these gases
+- Cardiac pulsation: heart rate trace
+- Respiration: change in chest volume
+
+.. _howtocollectphysdata:
+
+*How to collect each type of physiological data?*
```
smoia:
```suggestion
How to collect each type of physiological data?
```

docs/bestpractice.rst
```diff
@@ -4,4 +4,46 @@
...
+---------------------------------------------------
+
+- CO2 & O2: nasal cannula or face mask with gas analyzers, RespirAct (more specifically if you want to modulate end-tidal values)
+- Pulse: photo-plethysmograph, pulse oximeter
+- Respiration: respiratory belt, respiratory chest bellows
+
+.. _whattodowithphysdata:
+
+*What to do with physiological data once it has been collected?*
```
smoia:
```suggestion
What to do with physiological data once it has been collected?
```

Issue #128, opened 2020-01-17
If an AcqKowledge file is both in .txt and .acq format and you input the txt, the program reads the acq file instead


vinferrer:
## Expected Behavior
Reading txt file:
phys2bids -in Test_belt_pulse_samefreq.txt  -info -chtrig 2
File extension is .txt
Reading the file
phys2bids detected that your file is in AcqKnowledge format
Interval is not in seconds. Converting its value.
RESP - RSP100C PULSE - Custom, DA100C MR TRIGGER - Custom, HLT100C - A 5 PPG100C CO2 O2
Reading infos
File Test_belt_pulse_samefreq.txt contains:

00. RESP - RSP100C; sampled at 10000.0 Hz
01. MR TRIGGER - Custom, HLT100C - A 5; sampled at 10000.0 Hz
02. PPG100C; sampled at 10000.0 Hz
03. CO2; sampled at 10000.0 Hz
04. O2; sampled at 0.0 Hz
saving channels plot at plot at Test_belt_pulse_samefreq.png

## Steps to Reproduce the Problem

phys2bids -in phys2bids/phys2bids/tests/data/Test_belt_pulse_samefreq.txt  -info -chtrig 2
File extension is .acq
Reading the file
00. Channel RESP - RSP100C: 1336815 samples, 10000.0 samples/sec, loaded: True
01. Channel PULSE - Custom, DA100C: 1336815 samples, 10000.0 samples/sec, loaded: True
03. Channel PPG100C: 1336815 samples, 10000.0 samples/sec, loaded: True
04. Channel CO2: 1336815 samples, 10000.0 samples/sec, loaded: True
05. Channel O2: 1336815 samples, 10000.0 samples/sec, loaded: True
Reading infos
File phys2bids/phys2bids/tests/data/Test_belt_pulse_samefreq.acq contains:

00. RESP - RSP100C; sampled at 10000.0 Hz
01. PULSE - Custom, DA100C; sampled at 10000.0 Hz
02. PPG100C; sampled at 10000.0 Hz
03. CO2; sampled at 10000.0 Hz
04. O2; sampled at 10000.0 Hz
saving channels plot at plot at Test_belt_pulse_samefreq.png
## Specifications

    - Python version: Python 3.6.9
    - phys2bids version:phys2bids v1.2.0-beta+36.g24df90f.dirty
    - Platform: ubuntu

vinferrer:
[...]

PR #134, opened 2020-01-25
Adds logger

### LINKED ISSUE #28: Add logger

### LINKED ISSUE #142: Typos (+ small code suggestion)
## Typos
The command line output when calling phys2bids says:
`"saving channels plot at plot at my_phys_outputs/tutorial_file.png"`
This seems to be a typo, and should be something like:
`"saving channel plot to my_phys_outputs/tutorial_file.png"`

The command line output when calling phy2bids with the incorrect threshold says:
`Correcting time offset, assuming missing timepoints are at the beginning (try again with a more liberal thr`
Brackets are not closed, should be:
`Correcting time offset, assuming missing timepoints are at the beginning (try again with a more liberal thr)
`
## Suggestions

The help text in the doc string:
-chplot CHPLOT, --channels-plot CHPLOT
                        path to store channels plot

It was not immediately clear that it wanted the _full_ path, as in the directory and the filename. I actually think it would make more sense to make it only the directory (to match -indir and -outdir) and the name of the file should just be inputfilename_channelplot.png by default.

eurunuela:
Closes #28. Partially closes #142 .

## Proposed Changes

  - Adds logger and log file


eurunuela:
I believe this should cover all the logger statements. Please check I've not missed any.

smoia:
Can I ask someone else to review it? I think you need two reviewers anyway.
I have a couple of suggestions that i can send as comments anyway later on!

phys2bids/interfaces/txt.py
```diff
@@ -138,7 +143,7 @@ def process_acq(channel_list, chtrig, header=[]):
         freq = [freq] * (len(timeseries) + 1)
     else:
         if interval[-1].split('/')[0] != 'sec':
-            print('Interval is not in seconds. Converting its value.')
+            LGR.info('Interval is not in seconds. Converting its value.')
```
smoia:
```suggestion
            LGR.warning('Interval is not in seconds. Converting its value.')
```

phys2bids/physio_obj.py
```diff
@@ -268,34 +272,34 @@ def check_trigger_amount(self, thr=2.5, num_timepoints_expected=0, tr=0):
...
-            print('Checking number of timepoints')
+            LGR.info('Checking number of timepoints')
             if num_timepoints_found > num_timepoints_expected:
                 timepoints_extra = (num_timepoints_found
                                     - num_timepoints_expected)
-                print(f'Found {timepoints_extra} timepoints'
-                      ' more than expected!\n'
-                      'Assuming extra timepoints are at the end '
-                      '(try again with a more conservative thr)')
+                LGR.info(f'Found {timepoints_extra} timepoints'
```
smoia:
```suggestion
                LGR.warning(f'Found {timepoints_extra} timepoints'
```

phys2bids/physio_obj.py
```diff
@@ -268,34 +272,34 @@ def check_trigger_amount(self, thr=2.5, num_timepoints_expected=0, tr=0):
...
-                          ' are at the beginning (try again with '
-                          'a more liberal thr')
+                    LGR.warning('Correcting time offset, assuming missing timepoints'
+                                ' are at the beginning (try again with '
+                                'a more liberal thr')
                     time_offset -= (timepoints_missing * tr)
                 else:
-                    print('Can\'t correct time offset, (try again specifying '
-                          'tr or with a more liberal thr')
+                    LGR.error('Can\'t correct time offset, (try again specifying '
```
smoia:
```suggestion
                    LGR.warning('Can\'t correct time offset - you should specify the TR')
```

phys2bids/physio_obj.py
```diff
@@ -268,34 +272,34 @@ def check_trigger_amount(self, thr=2.5, num_timepoints_expected=0, tr=0):
...
+                              'tr or with a more liberal thr')
 
             else:
-                print('Found just the right amount of timepoints!')
+                LGR.info('Found just the right amount of timepoints!')
 
         else:
-            print('Cannot check the number of timepoints')
+            LGR.error('Cannot check the number of timepoints')
```
smoia: We could change this message to something like:
"you didn't specify the options necessary to perform this action".
If that's a warning, error or info I don't know.

phys2bids/utils.py
```diff
@@ -95,7 +98,7 @@ def check_input_type(filename, indir):
             break
 
     if fftype_found:
-        print(f'File extension is .{ftype}')
+        LGR.info(f'File extension is .{ftype}')
```
smoia: Here we could warn the users if they're looking for a particular extension txt) and we process another (acq).
Refers to #128 .

phys2bids/phys2bids.py
```diff
@@ -203,6 +207,31 @@ def _main(argv=None):
     outfile = os.path.join(options.outdir,
                            os.path.splitext(os.path.basename(options.filename))[0])
 
+    # Create logfile name
+    basename = 'phys2bids_'
+    extension = 'tsv'
+    isotime = datetime.datetime.now().replace(microsecond=0).isoformat()
```
smoia: Can we add the version number in the filename *and* as the first log info after execution?

eurunuela: I don't think the version number should go into the filename, that would make it unreadable. I'd say that the first log being the version number should be enough.

eurunuela:
Removing the `raise` statements made the tests fail. Will work on that.

phys2bids/phys2bids.py
smoia: Yeah, you're right.
However, I'm thinking about parallel computation - can we add the name of the input in the logger, or check that the same logger doesn't exists, and if so give it a different name?

eurunuela: This is an example of the log filename: `phys2bids_2020-02-11T19:46:08.tsv`. 

I'll add the input to the log file, it sounds like a good idea. However, I don't expect two log files to have the same name. See how the name even contains the second in which the command was executed. I don't think this will happen but if it did, I don't think they would be overwritten.

smoia: Alternatively, can we specify a destination for it? That way who wants to parallelise can save them in two different places.

eurunuela:
After fixing all the mess, it's ready to be reviewed again. I opted to not log the errors at the moment. We could add this feature in the future anyway as it is not that critical imo.

eurunuela:
Just when I thought it was ready, we have codecov (finally) working!

smoia:
...and ruining the party.
Codecov is reporting something about files that are not in the scope of this PR.
Let's close this one without worrying about it and then we can add more test to improve coverage.
Just worry about the reviewers :P

smoia:
Errata corrige: we need to test that the logger is doing its job. Do you want to add the test in this PR, or to open an issue referencing this one that calls for tests?

eurunuela:
I'll add the test and merge when #140 is merged.

vinferrer:
PR #140 merged, I see no reason to not merge this one

vinferrer:
oh yes, there is a conflict, damm

smoia:
And we miss tests!

eurunuela:
I'll take care of it!

eurunuela:
Decided I'll work on the test on a different PR.

eurunuela:
Okay, apparently the merging conflicts were not that clear. @vinferrer could you please update the `test_text.py` file to what it should be? You can do it directly in this PR.

smoia:
@all-contributors long time no see! Please add @eurunuela for code contributions!

smoia:
@all-contributors please add @eurunuela for code contributions!

smoia:
@all-contributors add @eurunuela for code

smoia:
Ok, doing it myself, got it. Don't be cranky you little bot.

eurunuela:
@smoia, regarding the automatic release thing, should this PR have both the enhancement and minormod labels? My guess is we can have more than one label in our PRs.

smoia:
Yes, you can have both Minormod and enhancement, but mainly because only Minormod will trigger auto, and both labels indicate an increase of version 0.+1.0.
You can refer to the contributor guidelines in the Auto PR to better understand the use of labels!

Issue #144, opened 2020-02-14
Add unit tests for the logger


eurunuela:
## Detailed Description

We should add testing for the changes introduced in #134 .

## Possible Implementation

I'd like to know what your opinion is on this matter as I'm still thinking of the best way to write tests fo the logger.

smoia:
We should probably check that the messages are reported - a bit like @vinferrer did in the `test_txt.py` test.

eurunuela:
Sorry, I didn't mean to close the issue. We could check the messages with an independent function in the integration tests.

smoia:
We can check it with a functional test, yes - let's do it like that!

eurunuela:
Already working on it!

PR #146, opened 2020-02-18
Adds integration tests

### LINKED ISSUE #130: Automatic testing on data with Travis
I believe we should have some automatic testing on data with Travis before we release our first non-beta version. This issue is related to #9.

## Context / Motivation

We should always make sure the code works on data, as it is the main purpose of the project. It would definitely make testing the code easier. It could even be helpful for the tutorials in our documentation.

## Possible Implementation

This feature should be implemented in a similar fashion to the unit tests.

vinferrer:
Closes #130 
## Proposed Changes

  - adds integration test to txt file
  -
  -


vinferrer:
This was unexpected, i thing the coverage increased that much because I am executing from the console to call the parser and therefore all phys2bids gets called

vinferrer:
OMG, this was unexpected. I think the coverage improves that much because i am running the phy2bids from the console and therefore all the suite gets executed

smoia:
That was expected, actually - and it was what we discussed in #130 and #144 (and was assigned to @eurunuela - please don't step on each others' feet!).
Link those issues and this PR, if you're making this become an integration or a function test - in which case, you need to add assertions on the outputs.
Also, testing cli/run.py doesn't necessarily mean we need to run the whole program.

vinferrer:
We decided to convert this PR in the integration test PR

vinferrer:
By the way, maybe we should separate these integration test in Travis from the normal unit test

eurunuela:
This is being taken care of by #166

PR #147, opened 2020-02-19
put number of samples in the header instead of the channel_list

vinferrer:
Closes #

## Proposed Changes

  - Now the sampling number is in the header not in the channel_list
  -
  -


vinferrer:
Very tiny change, once the test pass this is ready for review @smoia

PR #148, opened 2020-02-20
Trigger to 1

### LINKED ISSUE #118: Unify channel trigger numbering for txt and acq files
Summary
------------
So basically if the file is ``acq`` the first channel is 1 and if the file is``txt`` the first channel is 0.

Additional detail
--------------------
As example, if we start counting from 1, in Test_belt_pulse_multifreq.acq channel 2 is the trigger and therefore the instruction should be written as:
``phys2bids -in phys2bids/phys2bids/tests/data/Test_belt_pulse_multifreq.acq -info -chtrig 2``
whereas in Test_2minRest_trig_multifreq_header_no_comment.txt channel 2 is the trigger also but the instruction is written as:
``phys2bids -in phys2bids/phys2bids/tests/data/Test_2minRest_trig_multifreq_header_no_comment.txt -info -chtrig 1``

Next Steps
-------------
So we need to discuss if channel count starts at 1 or at zero and do the pertaining changes to the file reading functions

vinferrer:
Closes #118 

## Proposed Changes

  - Now channel indexing starts at 1 for the user in txt.py function


vinferrer:
@smoia review this

Issue #36, opened 2019-11-20
Chopping files semi-automatically 


RayStick:
A lot of researchers will have one physiological file for the whole scan session, which will need to be chopped into separate files corresponding to each scan to be analyzed. Offering a way of semi-automatically chopping these files, and allowing phys2bids to output multiple physiological files from the same input file, would be useful.

smoia:
This is a topic we briefly discussed at DCCC 2019 with @yarikoptic .
It would be a great enhancement!
However, we need to think about this in a smart way.
The way @yarikoptic was thinking about approaching it is by using timestamps. However, we don't have them by default in our files - so if interested in this option, we should definitely remember this for "good practices".

RayStick:
I have some MATLAB code that does this semi-automatically (it uses triggers for the time stamps alongside user input). At a later date, I could attempt to adapt this for phys2bids, if we wanted it.

RayStick:
An important related point to "chopping" files. I think there are two options:
1. The input file the user brings to phys2bids is already the part of the data that corresponds to one MRI scan (i.e. we don't offer a way of chopping files)
2. The input file the user brings to phys2bids can be a file that contains data from multiple scans, and phys2bids can chop it. _If_ it does chop it, we also have to consider the user may want the time stamp to start X seconds before the beginning of the scan and end X seconds after. This padding is important for later analysis.

smoia:
@tsalo might have an idea on how to do this!

tsalo:
I haven't had a chance to finalize it, but my approach with BioPac involves the following:

0. Set up your BioPac profile so that segments are labeled with a timestamp and a high maximum segment duration (3 hours)
1. When recording, have one long segment instead of stopping between scans.
2. Load data and identify trigger periods (onset and duration).
3. Infer absolute trigger period onsets from relative onsets (in samples), sample rate, and event marker label (the timestamp).
4. Load non-anonymized acquisition datetimes from the BIDS dataset.
5. Compare physio onsets with scan onsets to assign physio recordings to scans.
    - My guess is that this will require a little room for mismatch between the clock on the stimulus computer and the one on the physio computer.
6. Infer scan offsets based on scan onset, number of volumes, and repetition time.
    - Not sure how well this generalizes across scans folks might record physio for.
    - The reason I'm leaning toward using the scans to determine offsets instead of the physio trigger periods is because the physio trigger may not be disabled if the task is quit early. An alternative could be to use the heuristic, of course.
[...]

PR #150, opened 2020-02-21
Multifreq

### LINKED ISSUE #56: Add multifrequency support to labchart txt files
Once #54 is finish we have to add support for files with different frequency sampling in the recordings, we have to infer the frequency from the channels since labchart doesn't specify in the header

vinferrer:
Closes #56 

## Proposed Changes

  -Adds multi frequency support for labchart files
  -
  -


vinferrer:
unit test added closing #151

phys2bids/interfaces/txt.py
```diff
@@ -14,6 +14,34 @@
 LGR = logging.getLogger(__name__)
 
 
+def multifreq(timeseries, freq):
+    """
+    Checks if there are channels with different frequency than the maximum one
```
eurunuela: Could you please comment a bit more on what the function is doing?

phys2bids/interfaces/txt.py
```diff
@@ -14,6 +14,42 @@
...
+        # get value on the channel
+        for idx2, value in enumerate(chann[1:]):
+            # if value equal to previous value
+            if value == chann[idx2]:
+                # sample duplicated by interpolation
+                # increase counter so we have the number of
+                # equal samples continuos
+                eq_samples += 1
+            else:
+                break
```
smoia: @vinferrer , do you think we can check that the same amount of repeated samples is itself repeated twice or three times? Now, if for any reason a channel is perfectly 0 (say, the TR trigger) for the first 20 seconds, but then starts changing, we risk to save that channel with a different frequency than the real one. If we check that the "interpolation" is something recursive a couple of times, we're safe to assume is due to frequency resample.

phys2bids/interfaces/txt.py
```diff
@@ -14,6 +14,42 @@
 LGR = logging.getLogger(__name__)
 
 
+def multifreq(timeseries, freq):
```
smoia: Rather than "multifreq", can we call this something like "check_multifreq" or similar?

phys2bids/tests/test_txt.py
```diff
@@ -160,3 +160,15 @@ def test_raises():
     with raises(AttributeError) as errorinfo:
         txt.process_labchart(channels, chtrig, header)
     assert 'not in a valid LabChart' in str(errorinfo.value)
+
+
+def test_muiltifreq():
```
smoia:
```suggestion
def test_multifreq():
```

phys2bids/tests/test_txt.py
```diff
@@ -160,3 +160,15 @@ def test_raises():
...
+    # testing file already downloaded in the first test
```
smoia: I know the files are downloaded once already, but due to what happened during #145, I would rather download and delete the files for every test function - just to avoid problems.

vinferrer: That will significantly increase our testing time, but we can do it

phys2bids/interfaces/txt.py
vinferrer: sure

phys2bids/interfaces/txt.py
vinferrer: @smoia, last commit changed the function, now instead of checking only the begginning, the function asks for a start and a leftout. The function takes a part of the channels indicated by these parameters and in the while loop creates a list that contains how many times a number was repeated until changing. From this list the function counts the ocurrences and gets the most repeated number. This is the number that the frequency is divided by in the end. Theoretically the function could cover the hole channel, however I use the start and leftout arguments to make it faster.

vinferrer:
@smoia, last commit changed the function, now instead of checking only the begginning, the function asks for a start and a leftout. The function takes a part of the channels indicated by these parameters and in the while loop creates a list that contains how many times a number was repeated until changing. From this list the function counts the ocurrences and gets the most repeated number. This is the number that the frequency is divided by in the end. Theoretically the function could cover the hole channel, however I use the start and leftout arguments to make it faster.

smoia:
At the moment, this looks almost good to be merged for me.
@vinferrer can you adress @eurunuela's comment?
You also had doubts about the computational cost of checking more than one point of the channel, is it solved?

phys2bids/interfaces/txt.py
vinferrer: sure

phys2bids/interfaces/txt.py
vinferrer: resolved

phys2bids/tests/test_txt.py
vinferrer: done

phys2bids/tests/test_txt.py
vinferrer: changed

vinferrer:
No no. What i was saying is that checking the hole channel takes forever, that's why i leaved the start and leftout options so the function by default leaves out the beginning and the end of the channel. Right now the function is checking the hole channel excecpt for 1000 timepoints at the beggining and 1000 timepoints at the end

vinferrer:
Actually this makes me wonder if files with multiple sessions are gonna give problems

smoia:
I see. Let's think about it a bit more then. What about checking the channel at a random position around the middle and for a limited amount of points?

About the multisession files: One thing at a time: we still don't support them yet - that time will come (possibly with #36), and we'll think how to integrate things then.
For the moment, address the comments in the PR and let's try to merge it in.

eurunuela:
I would not check the channel at random positions. I think it would be wiser to check the start, middle and end of the channel unless we are certain that whatever random positions we choose the result will always be the same.

smoia:
It's actually more probable that the results will be the same checking three random positions in the middle rather than at the start and the end - in files that start earlier than the beginning of the sequence and have padding at the end, the trigger channel (and any other similar thing) might have the same value at the beginning and at the end for more than your sampling. This might lead to incorrect frequency classification.

smoia:
Solution: why don't we open an issue about it, and merge in this PR?

vinferrer:
We can discuss about this, I propose that the function does the process that is doing right now but in 3 subarrays that are seeded in random positions and put the size of these positions as an optional parameter

vinferrer:
but if you want we can merge this for the moment

vinferrer:
So, what do we do here?

smoia:
As soon as @eurunuela agrees on the changes, you can merge this in!

vinferrer:
let me solve the merge conflic

smoia:
However, please update your branch with respect to the master branch

vinferrer:
@smoia, @eurunuela I think with this last commit I solved the problem with the while loop. Now instead of limiting the while loop to a fix number, I get the maximum number of repeated samples as a reference, This way the number of samples that are not tested for multi frequency in  each channel is reduce to a few hundred but the hole array is inspected without need of dividing it. This means we don't have to open an issue about improving the multifrequency check

vinferrer:
therefore I think @smoia needs to check again the function

phys2bids/interfaces/txt.py
```diff
@@ -10,10 +10,64 @@
 import numpy as np
 
 from phys2bids.physio_obj import BlueprintInput
+from operator import itemgetter
+from collections import Counter
```
eurunuela: These should be added to the requirements if they're really necessary.

vinferrer:
Sry for running your approval

phys2bids/interfaces/txt.py
```diff
@@ -10,10 +10,64 @@
...
+    Parameters
+    ----------
+    timeseries: list
+        list with channels only in np array format
+    freq : list
+        list with the maximun frequency
+    start : integer
+        first sample of the channel to be considered
+    leftout : integer
+        number of sample at the end of the channel that are not considered
```
eurunuela:
```suggestion
        number of samples at the end of the channel that are not considered
```

phys2bids/interfaces/txt.py
```diff
@@ -10,10 +10,64 @@
...
+                    chann = chann[idx2 + 1:]
+                    if max_equal < eq_samples:
+                        max_equal = eq_samples
+                    break
+        # count the number of ocurrences in eq_list
+        dict_fr = Counter(eq_list)
+        # get maximum
+        n_inter_samples = max(dict_fr.items(), key=itemgetter(1))[0]
+        # if there are interpolated samples, it means the frequency is lower
+        # decrease frequency by dividing for the number of equal samples continuos
```
eurunuela: I do not understand this comment. Could you please elaborate a bit?

phys2bids/interfaces/txt.py
```diff
@@ -10,10 +10,64 @@
...
+        This is done  so this process doesn't take forever
+    Returns
+    -------
+    mfreq: list
+        new list with the actual frequency of the channels
+    """
+    mfreq = []
+    # for each channel check frequency
+    max_equal = 1
+    for idx, chann in enumerate(timeseries):
```
eurunuela: A for loop with a while loop inside that at the same time has another for loop inside worries me. This could end up being problematic. We should probably think of a better way of handling this, probably in a future update.

phys2bids/interfaces/txt.py
vinferrer: We can do that

phys2bids/interfaces/txt.py
vinferrer: We can think about this in the future but this was the easiest way to implement it

phys2bids/interfaces/txt.py
vinferrer: Basically the frequency is divided by n_inter_samples which is the number of samples that are interpolated

phys2bids/interfaces/txt.py
smoia: No they shouldn't, if they are core python packages

vinferrer: Are they?

phys2bids/interfaces/txt.py
smoia: Drop an issue about it!

vinferrer: As soon as we merge

phys2bids/interfaces/txt.py
smoia: That is something you need to tell us - did you install the packages? Can you check?

vinferrer: So i didn't install anything on travis, that should mean at least that the packages are required by the packages we have already as requisites

vinferrer: operator is a core module, and collections too you can found them here:
https://docs.python.org/3/library/

vinferrer: No need to install

smoia: Right - but you still need to check because if their support is dropped then we do need to add them to the requisites.
i quickly checked and they are core python modules. But please do recheck.

vinferrer: What do you want me to recheck?. I didn't installed them in travis and they are in the core library of python3

smoia: Yes don't worry - the message was posted after you posted your answer, but I couldn't see it due to the page refresh. You did recheck - it's good.

phys2bids/interfaces/txt.py
eurunuela: Cool! Thanks for the commit!

vinferrer:
Nice, Once @smoia finishes with the review I will merge

PR #152, opened 2020-02-25
solve list transpose that was buggy and problem with lines with extra blank tabs in reading

vinferrer:
## Proposed Changes

  - Solves reading problem when ``txt`` file has a line with extra tabs that creates null elements '', no instead of being an if, it's a while loop, therefore if there is more than one extra tab it deletes them all 
  - channel_list variables is a list of list of samplesxchannels and we want the transpose: channelsxsamples. Now we use  ``timeseries=list(map(list, zip(*channel_list)))`` instead of ``timeseries = np.matrix(channel_list).T.tolist()``. This affects acqknowledge and labchart files.

### DIFF
--- a/phys2bids/interfaces/txt.py
+++ b/phys2bids/interfaces/txt.py
@@ -80,7 +80,7 @@ def process_labchart(channel_list, chtrig, header=[]):
     orig_names_len = len(orig_names)
     names = ['time', 'trigger']
     # get channels
-    timeseries = np.matrix(channel_list).T.tolist()
+    timeseries = list(map(list, zip(*channel_list)))
     freq = [1 / interval[0]] * len(timeseries)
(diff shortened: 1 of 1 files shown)


phys2bids/interfaces/txt.py
```diff
@@ -80,7 +80,7 @@ def process_labchart(channel_list, chtrig, header=[]):
     orig_names_len = len(orig_names)
     names = ['time', 'trigger']
     # get channels
-    timeseries = np.matrix(channel_list).T.tolist()
+    timeseries = list(map(list, zip(*channel_list)))
```
eurunuela: Add a comment please.

vinferrer: done

PR #153, opened 2020-02-28
Trigger pictures improvement 

### LINKED ISSUE #15: Improving trigger pictures.
So far, the trigger picture created around [this line](https://github.com/smoia/phys2bids/blob/7dd17dded137fab2653b66812756ed967790dc30/physiobids.py#L409) have a simple title, and show the expected start and expected end of the sequence.
It would be nice to:
- [ ] add the name of the corresponding tsv output file in the title
- [ ] add a third subplot in the second row, that shows the last trigger
- [ ] rescale differently the Y axis in the first row, maybe the higher between twice the max value in the trigger array and three times the threshold?

vinferrer:
Closes #15 

## Proposed Changes

  - title shows tvs filename
  - second plot  in second row is plotted based on the index of the last point that exceeds the treshold
  - add a third subplot in the second row, that shows the last trigger
  - rescale differently the Y axis in the first row, maybe the higher between twice the max value in the trigger array and three times the threshold?


vinferrer:
image
plot as it is right now

vinferrer:
image
Current state of the plot

vinferrer:
@smoia this is how the plot looks like now:
image

vinferrer:
If you put it right:
image

vinferrer:
this is for the Test_belt_multifreq.txt

vinferrer:
However for the plots to be correct you have to input more parameters, look what happens for file 
Test2_time_trigger_CO2_O2_pulse_1000Hz_534TRs:
image

vinferrer:
The threshold is fine but nevertheless the second plots are wrong, I don't know if this could confuse the user

smoia:
Can you present it during the meeting next week? We can ask feedbacks then!

vinferrer:
Sure

vinferrer:
State of the picture with shadows:
Test_belt_pulse_samefreq_trigger_time

eurunuela:
Could we use another color that is softer and has less opacity? Something like #ffbb6e maybe?

vinferrer:
Test_belt_pulse_samefreq_trigger_time

eurunuela:
Maybe @danalclop has ideas on how to make these plots better looking?

vinferrer:
image
tutorial file plot

vinferrer:
image

eurunuela:
We should probably have the same range in the X axis of the plot on top as we do on the ones below. I mean, the one on the plot should probably go from -4 to 193 in this case.

vinferrer:
No we shouldn't the hole point of the plots below is knowing that you have entered the proper number of timepoints and the threshold for the program to find the first and las trigger

eurunuela:
Probably I wasn't very clear on my previous comment. Anyway, I've looked at the figure more thoroughly and think it's fine.

I do have a concern though, thinking from a user perspective. The last trigger is always centered except for the plot on the right on line 2. I'm aware of the different ways of plotting those figures. I'm just wondering if this could look weird for users.

vinferrer:
We can discuss it in the meeting

RayStick:
Do I wait for this to not be a draft, before giving a review?

eurunuela:
Yes @RayStick . It is not ready to review while it's a draft.

RayStick:
My main thoughts:
1. Looks great 
2. Adding a legend to explain the colours and lines would be helpful. 
3. Removing the final plot - for now, until issues #173 is resolved - makes sense.

vinferrer:
Figure after deleting suggested threshold plot and add legend

vinferrer:
image

vinferrer:
Maybe we should change the pictures of the documentation here too @smoia, @RayStick

smoia:
Sure, do you want to take care of it?

vinferrer:
I don't now where is it

vinferrer:
found them

smoia:
@vinferrer good! You might also want to change the description of the picture in this file:
https://github.com/physiopy/phys2bids/blob/master/docs/howto.rst

smoia:
@RayStick , can I ask you to have a look at the changes in the documentation once it's ready?

vinferrer:
Sure, so we still have to make the creation of the file only when -tr or -ntp are inputed and then, change the docs

vinferrer:
but wait, default tr is 1. ¿could that be problematic? my idea was that if the values are not the default then plot

eurunuela:
Is the default TR used somewhere else in the code?

phys2bids/phys2bids.py
```diff
@@ -265,15 +265,19 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
...
-    LGR.info('Plot trigger')
-    plot_path = os.path.join(outdir,
-                             os.path.splitext(os.path.basename(filename))[0])
-    if sub:
-        plot_path += f'_sub-{sub}'
-    if ses:
-        plot_path += f'_ses-{ses}'
-    viz.plot_trigger(phys_in.timeseries[0], phys_in.timeseries[chtrig],
-                     plot_path, tr, thr, num_timepoints_expected)
+    if tr != 1 or num_timepoints_expected != 0:
```
eurunuela: What if the TR is actually 1?

vinferrer: I think is highly unlikely and also the moment you add the -npt i won't matter. In the logger it is actually recommended to add both

vinferrer: But if you guys think if can become an issue, we can use only -ntp

smoia: Why don't we set the default TR (in the CLI) to 0 instead?

vinferrer: good idea, let me change it

vinferrer: I am gonna check is not a problem in other things, if we introduce it as zero

vinferrer: the program doesn't stop with tr=0 as default, let me push to check travis

smoia: The whole program shouldn't stop with default TR, be it one or zero!

vinferrer: seems like test are fine, I think this conversation can be resolved

RayStick:
@smoia yep! But it may be Sunday or Monday that I can review

vinferrer:
changes to docfile done, you can start the review

vinferrer:
@RayStick, @smoia

phys2bids/phys2bids.py
```diff
@@ -265,15 +265,19 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
...
+        plot_path = os.path.join(outdir,
+                                 os.path.splitext(os.path.basename(filename))[0])
+        if sub:
+            plot_path += f'_sub-{sub}'
+        if ses:
+            plot_path += f'_ses-{ses}'
+        viz.plot_trigger(phys_in.timeseries[0], phys_in.timeseries[chtrig],
+                         plot_path, tr, thr, num_timepoints_expected, filename)
+    else:
+        LGR.info('Not plotting trigger. If you want trigger to be plot please'
```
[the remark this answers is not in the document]
vinferrer: done

docs/howto.rst
```diff
@@ -118,7 +118,7 @@ This is outputted to the terminal:
     INFO:phys2bids.viz:saving channel plot to tutorial_file.png
     INFO:phys2bids.physio_obj:Counting trigger points
     WARNING:phys2bids.physio_obj:The necessary options to find the amount of timepoints were not provided.
-    INFO:phys2bids.phys2bids:Plot trigger
+    INFO:phys2bids.phys2bids:Not plotting trigger. If you want trigger to be plot pleaseplease enter -tr or -ntp, preferebly both
```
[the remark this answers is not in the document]
vinferrer: done

phys2bids/viz.py
```diff
@@ -47,24 +49,67 @@ def time2ntr(x):
...
+                    labelbottom=False,
+                    pad=15)
+    # add secondary axis ticks
     subplot.secondary_xaxis('top', functions=(time2ntr, ntr2time))
-    subplot.plot(time, trigger, '-', time, time, '-')
+    # add secondary axis labelS
+    ax2.set_xlabel('TR')
+    subplot.plot(time, trigger, '-', time, block, '-')
+    subplot.fill_between(time, block, where=block >= d, interpolate=True, color='#ffbb6e')
+    ax2.set_title('Initial trigger for selected threshold')
```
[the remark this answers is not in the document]
vinferrer: done

phys2bids/viz.py
```diff
@@ -47,24 +49,67 @@ def time2ntr(x):
...
+                    axis='x',          # changes apply to the x-axis
+                    which='both',      # both major and minor ticks are affected
+                    bottom=False,      # ticks along the bottom edge are off
+                    top=False,         # ticks along the top edge are off
+                    labelbottom=False,
+                    pad=15)
     subplot.secondary_xaxis('top', functions=(time2ntr, ntr2time))
-    subplot.plot(time, trigger, '-', time, time, '-')
+    ax2.set_xlabel('TR')
+    ax2.set_title('Last trigger for selected threshold')
```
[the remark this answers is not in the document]
vinferrer: done

vinferrer:
@RayStick Changes done, you can check them and if there is no more changes, approve the request.

Cheers

PR #155, opened 2020-02-28
Create only time channel from trigger channel, don't move around trigger

smoia:
Closes #

To be reviewed after #154 

## Proposed Changes

  - Now interfaces.acq.py reads the file without moving around the supposed trigger channel. It only creates the time from it.


smoia:
I'm going to close this and reopen as a draft as it's a bit more refactoring than imagined.

PR #156, opened 2020-02-28
Stop reordering trigger

smoia:
Closes nothing
Addresses #118 and possibly #148 .

## Proposed Changes

  - Trigger isn't moved any more to channel 1, instead it's kept where it should be.


vinferrer:
Should we address this also for txt files here

vinferrer:
I can tell you the needed changes, it should be only a few lines

smoia:
Let's work on this together, possibly we'll need to merge #148 with this?

smoia:
I don't know if this should be a refactoring and majormod PR, or a BugFIX PR. I added all the labels in case, but let me know what you think and remove the label(s) that are not necessary.

I checked the drop in testing, and it's due to the fact that `physio_obj.py` and `phys2bids.py` are not tested yet. I would pass them through integration tests, but if not, I'll open a new issue for tests (and eventually a PR).

PR #165, opened 2020-03-06
Adjusts main function for easier testing and usage of phys2bids from inside python

eurunuela:
Closes #

## Proposed Changes

  - Adjusts main function for easier testing and usage of phys2bids from inside python.
  - Basically, the main function called phys2bids now receives arguments from the parser instead of calling the parser itself.


smoia:
We're going to increase coverage after this PR, so codecov reports can be ignored!

PR #166, opened 2020-03-06
Adds integration tests

### LINKED ISSUE #130: Automatic testing on data with Travis

eurunuela:
Closes #130 and #144 

## Proposed Changes

  - Adds integration test for tutorial file.
  - Tests the logger.
  - Integration tests for more data types will be added soon.


eurunuela:
@rmarkello any idea on why the only test that fails is the wheel one? Can't find anything related to this online...

rmarkello:
**Short answer:** Looks like this line is causing the issue:

https://github.com/physiopy/phys2bids/blob/8aa8e3539ad18fc3afa3f41f560088c989bef2a8/.travis.yml#L34

Get rid of that and you *should* be good-to-go.

**Long answer:** Because you add the `phys2bids` git directory (with the above line) to the `$PYTHONPATH` when you try and install via the wheel, `setuptools` detects an "existing" installation of the module (it's on your `$PYTHONPATH`, after all!) so it doesn't actually install the package. Pip tries to warn you about this on line 377 of the Travis logs:

```
Requirement already satisfied: phys2bids==1.3.0b0+146.g422cf2d from file:///home/travis/build/physiopy/phys2bids/dist/phys2bids-1.3.0b0%2B146.g422cf2d-py3-none-any.whl in /home/travis/build/physiopy/phys2bids (1.3.0b0+146.g422cf2d)
```

Because you never officially install the package, the `phys2bids` CLI isn't available (see line 460 of the Travis log)&mdash;but you can still *import* it, so the rest of the tests pass. (The reason this only happens with the wheel is because the other installation methods don't run the same "is it installed already?" check beforehand.)

If you had used something like `subprocess.run` instead of `os.system` to run the CLI I think the test would have errored out a bit earlier than it did. Either way, if you remove that line from the `.travis.yml` then I think it should work either way, but you might want to consider using `subprocess.run` all the same&mdash;it's a bit more robust than `os.system`!

I'm gonna assign myself as a reviewer for this PR and will take a look at it again once you've got everything passing!

Let me know if you have any questions in the interim :smile:

rmarkello:
Ah, I think this is gonna fail because of your call to `subprocess.run` unfortunately... It's not exactly the same function call as `os.system`. Sorry for not giving you a heads up about that! :see_no_evil:

Try adding these two parameters to the function calls:

```python
subprocess.run(command, shell=True, check=True)
```

The `shell=True` lets you pass a string as the command (instead of a list of arguments), and the `check=True` will error if the sub-process command didn't exit cleanly.

smoia:
@rmarkello saves the day!

eurunuela:
Wow, really appreciate your quick and insightful response @rmarkello , I'm still learning this whole Travis thing.

Regarding the `subprocess.run` command, I did think of using it but thought it was not worth the extra import. Apparently it is, so good to know!

rmarkello:
Wait for the tests to pass to say that, @smoia ! :laughing: 

@eurunuela: If it's a standard library import I don't think you should be shy about using it. If you'd be adding an extra *dependency* then I think it's worth considering, but as-is I think `subprocess` is the way to go! :+1:

eurunuela:
Well, safe to say you saved the day @rmarkello ! 🎉

eurunuela:
I'll add the ACQ and multi-frequency tests before making the PR ready for review. I'll try to have it tomorrow morning.

eurunuela:
I was wondering whether we should add a test for heuristics in a multi-frequency setting. What do you guys think?

Otherwise, I think this PR is ready.

eurunuela:
@rmarkello any chance on reviewing before our meeting tomorrow?

rmarkello:
I will work on this soon! Sorry, been in meetings all day. Will definitely get to it before the meeting tomorrow 👍

eurunuela:
@rmarkello don't worry! We're all very busy lately. 

I just made sure that the last addition to the tests works. All yours now! 😉

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+                   f'-chtrig {test_chtrig} -outdir {test_outdir}', shell=True, check=True)
+
+    # Read logger file
+    files = os.listdir(test_path)
+    logger_idx = [log_idx for log_idx, log_str in enumerate(
+                      files) if 'phys2bids' in log_str]
+    logger_file = files[logger_idx[0]]
+
+    with open(os.path.join(test_path, logger_file)) as logger_info:
+        logger_info = logger_info.readlines()
```
rmarkello: Can we reduce this code a bit using e.g., a `glob`?

```suggestion
    # Read logger file
    logger_file = glob.glob(os.path.join(test_path, '*phys2bids*'))[0]

    with open(os.path.join(test_path, logger_file)) as logger_info:
        logger_info = logger_info.readlines()
```

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    test_tr = 1.2
+    test_thr = 0.735
+    phys2bids(filename=test_full_path, chtrig=test_chtrig, outdir=test_outdir,
+              num_timepoints_expected=test_ntp, tr=test_tr, thr=test_thr)
+
+    # Check that files are generated
+    assert os.path.isfile(os.path.join(test_path, 'tutorial_file.log'))
+    assert os.path.isfile(os.path.join(test_path, 'tutorial_file.json'))
+    assert os.path.isfile(os.path.join(test_path, 'tutorial_file.tsv.gz'))
+    assert os.path.isfile(os.path.join(test_path, 'tutorial_file_trigger_time.png'))
```
rmarkello: What about a for-loop here? E.g.,

```suggestion
    # Check that files are generated
    for suffix in ['.log', '.json', '.tsv.gz', '_trigger_time.png']:
        assert os.path.isfile(os.path.join(test_path, 'tutorial_file' + suffix))
```

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+
+    # Read log file (note that this file is not the logger file)
+    with open(os.path.join(test_path, 'tutorial_file.log')) as log_info:
+        log_info = log_info.readlines()
+
+    # Check timepoints expected
+    expected_idx = [log_idx for log_idx, log_str in enumerate(
+                      log_info) if 'Timepoints expected' in log_str]
+    expected_found = log_info[expected_idx[0]]
+    assert '158' in expected_found
```
rmarkello: Since you're using a really consistent format here (e.g., checking for a string in the log file), can you make this (along with the rest of the checks) a big for-loop? That way you're only iterating through the log text once (instead of e.g., four times), and it reduces the repetition a bit. Something like:

```python
for log_idx, log_str in enumerate(log_info):
    for text, expected in [('Timepoints expected', '158'), 
                           ('Sampling Frequency', '1000.0'),
                           ('Sampling started', '0.24499999'),
                           ('first trigger', 'Time 0')]:
        if text in log_str:
            assert expected in log_str
```

Or, you could make a helper function!

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+
+    # Check timepoints found
+    timepoints_idx = [log_idx for log_idx, log_str in enumerate(
+                      log_info) if 'Timepoints expected' in log_str]
+    timepoints_found = log_info[timepoints_idx[0]]
+    assert '158' in timepoints_found
+
```
rmarkello:
```suggestion
```

This is a repetition of the code above, from what I can tell, so I think it's safe to remove!

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    assert 'Time 0' in start_found
+
+    # Checks json file
+    # json_data = json.load(os.path.join(test_path, 'tutorial_file.json'))
+    with open(os.path.join(test_path, 'tutorial_file.json')) as json_file:
+        json_data = json.load(json_file)
+
+    # Compares values in json file with ground truth
+    assert json_data['SamplingFrequency'] == 1000.0
+    assert json_data['StartTime'] == 0.24499999999989086
```
rmarkello: This isn't super safe... It might be worth using e.g., `math.isclose` instead!

```suggestion
    assert math.isclose(json['StartTime'], 0.245)
```

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    test_chtrig = 3
+
+    phys2bids(filename=test_filename, indir=test_path, outdir=test_path,
+              chtrig=test_chtrig, num_timepoints_expected=1)
+
+    # Check that files are generated
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_samefreq.log'))
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_samefreq.json'))
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_samefreq.tsv.gz'))
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_samefreq_trigger_time.png'))
```
rmarkello: Same comment as the above function regarding potentially making this a for-loop!

phys2bids/tests/test_integration.py
rmarkello: Is this supposed to be 'Timepoints Found' instead? Looking at the below tests which follow a similar structure suggests it might be!

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+
+    # Read log file (note that this file is not the logger file)
+    with open(os.path.join(test_path, 'Test_belt_pulse_samefreq.log')) as log_info:
+        log_info = log_info.readlines()
+
+    # Check timepoints expected
+    expected_idx = [log_idx for log_idx, log_str in enumerate(
+                      log_info) if 'Timepoints expected' in log_str]
+    expected_found = log_info[expected_idx[0]]
+    assert '1' in expected_found
```
rmarkello: As in the above test, it might be worth making this (and the below checks) a for-loop, or creating that helper function or something!

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    start_found = log_info[start_idx[0]]
+    assert 'Time 0' in start_found
+
+    # Checks json file
+    with open(os.path.join(test_path, 'Test_belt_pulse_samefreq.json')) as json_file:
+        json_data = json.load(json_file)
+
+    # Compares values in json file with ground truth
+    assert json_data['SamplingFrequency'] == 10000.0
+    assert json_data['StartTime'] == 10.425007798392297
```
rmarkello:
```suggestion
    assert math.isclose(json_data['StartTime'], 10.425007798392297)
```

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+              chtrig=test_chtrig, num_timepoints_expected=1)
+
+    # Check that files are generated
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_multifreq_625.0.log'))
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_multifreq_625.0.json'))
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_multifreq_625.0.tsv.gz'))
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_multifreq_trigger_time.png'))
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_multifreq_10000.0.log'))
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_multifreq_10000.0.json'))
+    assert os.path.isfile(os.path.join(test_path, 'Test_belt_pulse_multifreq_10000.0.tsv.gz'))
```
rmarkello: Same comment: could we for-loop this?

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    """
+    # Read log file of frequency 625 (note that this file is not the logger file)
+    with open(os.path.join(test_path, 'Test_belt_pulse_multifreq_625.0.log')) as log_info:
+        log_info = log_info.readlines()
+
+    # Check timepoints expected
+    expected_idx = [log_idx for log_idx, log_str in enumerate(
+                      log_info) if 'Timepoints expected' in log_str]
+    expected_found = log_info[expected_idx[0]]
+    assert '1' in expected_found
```
rmarkello: Same comment as above tests: creating a helper function or making this a for-loop would be great!

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    start_found = log_info[start_idx[0]]
+    assert 'Time 0' in start_found
+
+    # Checks json file
+    with open(os.path.join(test_path, 'Test_belt_pulse_multifreq_625.0.json')) as json_file:
+        json_data = json.load(json_file)
+
+    # Compares values in json file with ground truth
+    assert json_data['SamplingFrequency'] == 625.0
+    assert json_data['StartTime'] == 0.29052734375
```
rmarkello:
```suggestion
    assert math.isclose(json_data['StartTime'], 0.29052734375)
```

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    start_found = log_info[start_idx[0]]
+    assert 'Time 0' in start_found
+
+    # Checks json file
+    with open(os.path.join(test_path, 'Test_belt_pulse_multifreq_10000.0.json')) as json_file:
+        json_data = json.load(json_file)
+
+    # Compares values in json file with ground truth
+    assert json_data['SamplingFrequency'] == 10000.0
+    assert json_data['StartTime'] == 10.425007798392297
```
rmarkello:
```suggestion
    assert math.isclose(json_data['StartTime'], 10.425007798392297)
```

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    assert 'Time 0' in start_found
+
+    # Checks json file
+    json_filename = 'sub-006_ses-01_task-test_rec-labchart_run-00_physio.json'
+    with open(os.path.join(test_path_output, json_filename)) as json_file:
+        json_data = json.load(json_file)
+
+    # Compares values in json file with ground truth
+    assert json_data['SamplingFrequency'] == 1000.0
+    assert json_data['StartTime'] == 0.24499999999989086
```
rmarkello:
```suggestion
    assert math.isclose(json_data['StartTime'], 0.24499999999989086)
```

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    """
+    # Read log file of frequency 625 (note that this file is not the logger file)
+    with open(os.path.join(test_path, 'Test_belt_pulse_multifreq_10000.0.log')) as log_info:
+        log_info = log_info.readlines()
+
+    # Check timepoints expected
+    expected_idx = [log_idx for log_idx, log_str in enumerate(
+                      log_info) if 'Timepoints expected' in log_str]
+    expected_found = log_info[expected_idx[0]]
+    assert '1' in expected_found
```
rmarkello: Either the for-loop or helper function for this and the checks below, too!

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+
+    test_path_output = os.path.join(test_path, 'sub-006/ses-01/func')
+
+    # Check that files are generated
+    assert os.path.isfile(os.path.join(test_path_output, ('sub-006_ses-01_task-test_rec-'
+                                                          'labchart_run-00_physio.json')))
+    assert os.path.isfile(os.path.join(test_path_output, ('sub-006_ses-01_task-test_rec-'
+                                                          'labchart_run-00_physio.log')))
+    assert os.path.isfile(os.path.join(test_path_output, ('sub-006_ses-01_task-test_rec-'
+                                                          'labchart_run-00_physio.tsv.gz')))
```
rmarkello: For-loop to check these files?

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    # Read log file (note that this file is not the logger file)
+    log_filename = 'sub-006_ses-01_task-test_rec-labchart_run-00_physio.log'
+    with open(os.path.join(test_path_output, log_filename)) as log_info:
+        log_info = log_info.readlines()
+
+    # Check timepoints expected
+    expected_idx = [log_idx for log_idx, log_str in enumerate(
+                      log_info) if 'Timepoints expected' in log_str]
+    expected_found = log_info[expected_idx[0]]
+    assert '158' in expected_found
```
rmarkello: For-loop or helper function to check these expected outputs!

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+    subprocess.run(command_str, shell=True, check=True)
+
+    # Check that plot all file is generated
+    assert os.path.isfile('tutorial_file.png')
+
+    # Read logger file
+    files = os.listdir(test_path)
+    logger_idx = [log_idx for log_idx, log_str in enumerate(
+                      files) if 'phys2bids' in log_str]
+    logger_file = files[logger_idx[0]]
```
rmarkello: Replacing this with a `glob.glob` call would be great!

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,466 @@
...
+
+    # Check files were correctly read
+    with open(os.path.join(test_path, logger_file)) as logger_info:
+        logger_info = logger_info.readlines()
+
+    # Get trigger info
+    trigger_idx = [log_idx for log_idx, log_str in enumerate(
+                      logger_info) if 'Trigger; sampled at' in log_str]
+    trigger_found = logger_info[trigger_idx[0]]
+    assert '1000.0' in trigger_found
```
rmarkello: Either the for-loop or helper function for checking these outputs would be ideal!

phys2bids/tests/test_integration.py
eurunuela: Didn't know about this function, thanks!

phys2bids/tests/test_integration.py
eurunuela: You're right! Copy pasting error. It should be 'Timepoints Found'.

phys2bids/tests/test_integration.py
eurunuela: Good idea thanks! I didn't even think of it as I was writing the tests as I thought of them. This will surely make the tests easier to read.

eurunuela:
@rmarkello thanks a lot for your feedback! I think it really has made the tests much easier to read and more robust. I think we're ready to merge. What do you think?

phys2bids/tests/test_integration.py
```diff
@@ -0,0 +1,348 @@
...
+        logger_info = logger_info.readlines()
+
+    # Get version info
+    version_idx = [log_idx for log_idx, log_str in enumerate(
+                      logger_info) if 'phys2bids version' in log_str]
+    version_found = logger_info[version_idx[0]]
+    current_version = get_versions()
+
+    # Checks logger version is the current version
+    assert current_version['version'] in version_found
```
rmarkello:
```suggestion
    # Get version info
    current_version = get_versions()
    assert check_string(logger_info, 'phys2bids version', current_version)
```

PR #170, opened 2020-03-10
Add files for new logo & favicon

danalclop:
Closes #

## Proposed Changes

 - raw Pixelmator file: 'phys2BIDSlogo.pxm';
 - updated version of the logo (without folders) in .PNG: 'phys2bids_logo2_1280×640.png';
 - negative version of the updated logo for the Read the Docs page: 'phys2bids_negativeReadTheDocs_500x150.png';
 - two versions of the favicon (with and without background) for the tabs: 'favicon.zip';


eurunuela:
I've checked the failing test and it doesn't make sense. This PR could be merged regardless of the failing test as it does not interfere. However, I'd say @smoia should review it too.

smoia:
I don't think the errors in Travis are related to this PR, given that we deal with heavier files and the error seems to happen due to OSF. Can you quickly check that the URL inside the tests reflects the URL of the files in storage in OSF?

The new logos are for the whole organisation, "physiopy", not "phy2bids", so if you could rename the files with the new logos "physiopy" and produce the negative of "phys2bids" as it is now, it would be great!

eurunuela:
Both urls correspond to the files being downloaded in the test.

This is not an issue we saw on previous PRs. Could it be a Travis issue?

PR #176, opened 2020-03-13
Fix/logger

### LINKED ISSUE #171: OSError on windows 10
**When trying to run the [test illustrated in the docs](https://phys2bids.readthedocs.io/en/latest/howto.html) from a windows environment, I get an OS error. Even though the package is at an early stage, I wanted to mention it because the problem seems simple to solve.**

## Expected Behavior
When in directory `phys2bids/tests/data`, and running the command `phys2bids -in Test_multiscan100Hz_trig_samefreq_header.txt -info`, you would expect a nice output like the one in the docs.

## Actual Behavior
Instead of the beautiful output, I get that 
image

From my understanding, it seems like the *Logging* package that is used to produce the output file name doesn't spit a proper path. On Windows, the directory looks somewhat like this `C:\Users\etc\etc`. 

Python usually converts these windows path to either a string format like this : `'C:\Users\etc\etc'` 

or a double back-slash format like this : `C:\\Users\\etc\\etc`

In my case, it seems like it does both, so it's considered an invalid argument : `'C:\\Users\\etc\\etc'`

**Everything runs smoothly on a Linux/Ubuntu 18.04LTS system**

## Steps to Reproduce the Problem

1.  Go on a `windows/latest` system
2.  Have your python env installed with phys2bids (On my machine, **GitBash** is installed using**Linux Subsystem for Windows**. and my Conda environment is specified in $PATH. it doesn't look like the usual windows command line interface).
3. Open Anaconda prompt, or a Command Line application fed with your conda path.
4. Reproduce steps given in the [docs](https://phys2bids.readthedocs.io/en/latest/howto.html)

## Specifications

    - Python version: 3.7
    - phys2bids version: dev version, `phys2bids 1.3.0b0+111.g20fc77e`
    - Platform: 2020-03 Cumulative Update for Windows 10 Version 1809 for x64-based Systems (KB4538461)

## Possible solution
So, I'm not a expert, but it seems like the problem lies in the way the output path is formatted here (Line 211, `phys2bids.py` : 

   ``` # Set up logging file and open it for writing
    log_handler = logging.FileHandler(logname)
    log_handler.setFormatter(log_formatter)
    sh = logging.StreamHandler() 
```

**I've test all functions and it systematically gives this error. I suggest that in further versions, we include a conditional statement for windows path format** (i.e. if windows users want to format their phys files in BIDS).  I will post my attempts to solve this issue here

### LINKED ISSUE #174: Add Windows CI testing
We are currently not testing (nor developing) `phys2bids` on Windows systems, leading to issues like e.g., #171. I think we should look into adding Windows CI testing using a service like [AppVeyor](https://www.appveyor.com/) or [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/) (since TravisCI currently only has [minimal](https://docs.travis-ci.com/user/reference/windows/) Windows support).

## Context / Motivation
We shouldn't limit our users to those researchers working on Mac or Linux systems, so adding AppVeyor or Azure support / testing so that we can ensure our primary workflows are viable on Windows is critical.

## Possible Implementation
`nipy/nibabel` recently moved to Azure Pipelines from AppVeyor (relevant files: [1](https://github.com/nipy/nibabel/blob/master/.azure-pipelines/windows.yml) [2](https://github.com/nipy/nibabel/blob/master/azure-pipelines.yml)); we could look into simply copying the relevant portions of their framework for `phys2bids`.

eurunuela:
Closes #171.

## Proposed Changes

  - Fixes logger bug on Windows.

### DIFF
--- a/phys2bids/phys2bids.py
+++ b/phys2bids/phys2bids.py
@@ -199,7 +199,7 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     # Create logfile name
     basename = 'phys2bids_'
     extension = 'tsv'
-    isotime = datetime.datetime.now().replace(microsecond=0).isoformat()
+    isotime = datetime.datetime.now().strftime('%Y-%m-%dT%H%M%S')
     logname = os.path.join(outdir, (basename + isotime + '.' + extension))
 
     # Set logging format
--- a/phys2bids/tests/test_integration.py
+++ b/phys2bids/tests/test_integration.py
@@ -48,7 +48,7 @@ def test_logger():
 
     # Get version info
     current_version = get_versions()
-    assert check_string(logger_info, 'phys2bids version', current_version)
+    assert check_string(logger_info, 'phys2bids version', current_version['version'], is_num=False)
 
     # Removes generated files
     os.remove(os.path.join(test_path, logger_file))


eurunuela:
Thanks for the quick review @rmarkello !

To be honest, I don't know why the changes to `test_integration.py` were added into this PR.

sangfrois:
**Everything runs smoothly on my windows machine! Excellent work guys!!**

eurunuela:
I’m glad it works @sangfrois !

Issue #162, opened 2020-03-05
Improve check multifreq function


vinferrer:
## Summary
the ````check_multifreq`` function inside ``interfaces.txt```has  a for loop with a while loop inside that at the same time has another for loop inside. It would be better if we had a different implementation:

## Next Steps

    *  propose alternatives
    *  Create PR to solve this
    *

PR #178, opened 2020-03-18
Txt_test_faster

### LINKED ISSUE #172: Make txt test faster
## Detailed Description

Test of txt files are very slow due to the amount of downloading and reading of files that happen in each of the testing functions.

## Context / Motivation

Making tests faster would definitely be beneficial for the project.

## Possible Implementation

We could write a helper function that takes care of the downloading of the data and assigns it to variables that are then passed into the testing functions.

vinferrer:
Closes #172 

## Proposed Changes

  - Now the tested ``acq`` file has only the trigger and other channel and the length of the channels is 1048560
  - All the downloads are done at the beggining and then the functions get the needed parameters. The only exception is ``test_populate_phys_input`` because the tested function has to read and this was giving problems if the file wasn't download inside the function
- Testing reduced from 7 to 5 min


rmarkello:
Hey @vinferrer! Thanks so much for getting this started&mdash;I'm really glad to hear that the testing time is down ~25%!

My primary concern about this new implementation is that it might lead to difficulties detecting bugs in the codebase. That is, if there's an issue in any of the code that you've moved outside functions (e.g., if a file is accidentally moved or deleted from OSF, or there's an issue in the `txt.read_header_and_channels()` function for one of the test files), `pytest` will error during test _collection_ rather than during the tests themselves. This makes debugging significantly harder&mdash;we don't get the helpful information `pytest` tries to give us about what tests failed / passed. (It will tell us what file / line caused the error during collection, but the whole testing suite aborts before _any_ tests are run. You can see what this is like by modifying one of the OSF URLs in the download section of `test_txt.py` right now (e.g., change `'https://osf.io/4yudk/download'` to `'https://osf.io/INVALID/download'`)  and then running `pytest` on everything.)

One potential solution (as I described in #172) would be to use `pytest.fixtures` and add discrete download functions. When you do that, `pytest` is pretty verbose about exactly where the error is, and the entire test suite will run (even if some of the tests /fixtures fail). I'm happy to discuss alternatives or implementation specifics in more detail, but I think the current set-up (with much of the code living "freeform" inside the testing modules) isn't ideal.

rmarkello:
Since I haven't done much codestuff for `phys2bids` in a while would you be willing for me to maybe make a PR on your PR, implementing some of my suggestions? Let me know if you'd be open to that 😄

vinferrer:
I see, let me code a bit tomorrow, I want a try do it by my self

vinferrer:
> 
> 
> LGTM! 2 minutes faster is quite a thing! Do you know which of the tests is taking the longest to run?
> 
> I think that #162 could help with reducing testing time even more. What do you think @vinferrer ?

That we can treat in another PR, but sure reducing check_multifreq time would be great

vinferrer:
> 
> 
> Hey @vinferrer! Thanks so much for getting this started—I'm really glad to hear that the testing time is down ~25%!
> 
> My primary concern about this new implementation is that it might lead to difficulties detecting bugs in the codebase. That is, if there's an issue in any of the code that you've moved outside functions (e.g., if a file is accidentally moved or deleted from OSF, or there's an issue in the `txt.read_header_and_channels()` function for one of the test files), `pytest` will error during test _collection_ rather than during the tests themselves. This makes debugging significantly harder—we don't get the helpful information `pytest` tries to give us about what tests failed / passed. (It will tell us what file / line caused the error during collection, but the whole testing suite aborts before _any_ tests are run. You can see what this is like by modifying one of the OSF URLs in the download section of `test_txt.py` right now (e.g., change `'https://osf.io/4yudk/download'` to `'https://osf.io/INVALID/download'`) and then running `pytest` on everything.)
> 
> One potential solution (as I described in #172) would be to use `pytest.fixtures` and add discrete download functions. When you do that, `pytest` is pretty verbose about exactly where the error is, and the entire test suite will run (even if some of the tests /fixtures fail). I'm happy to discuss alternatives or implementation specifics in more detail, but I think the current set-up (with much of the code living "freeform" inside the testing modules) isn't ideal.

Talking about this, should we use the pytest.fixtures functions for dowloading or can we add also the common preproccessing before the asserts?

vinferrer:
Okay guys I refactored the code and now I am using ``@pytest.fixture`` the testing time is more or less the same (5 min) except if codecov gets executed, then is 7. I think you can do the review now @rmarkello.

smoia:
@rmarkello, there is no rush in merging this in, if you want to make some suggestions!
However, if it's about testing in general, I would ask you to open another issue/PR/both (and merge this in, please!).

PR #180, opened 2020-03-19
[TEST] Consolidates / updates test suite

### LINKED ISSUE #179: Consolidate testing suite
## Summary
With @vinferrer's updates to the test suite (via #178) merged (🙌), I thought it might be worth re-visiting the test suite in a more general sense. It's evolved pretty rapidly as development has happened on `phys2bids` and I think it might be worth doing a bit of consolidation. I wanted to make this issue to discuss ways that we might do that, and offer a few initial suggestions. (I will also open a PR shortly after posting this issue implementing some of these suggestions to give a better idea of what this would look like in practice.)

### 1. Testing files

First, I wanted to determine what the differentiation is between the files that are included in the `phys2bids/tests/data` directory versus the files that are on OSF and are downloaded in the test suite. Should we just move all the files in that directory onto OSF, or is there some reason they're included? For example, I can't find the `Test_belt_pulse_multifreq.acq` or `Test_multiscan100Hz_trig_samefreq_header.txt` that are in that directory used anywhere in `phys2bids`, suggesting they should possibly be removed.

I understand wanting to have some files for the documentation (e.g., `tutorial_file.{txt,json}`), but I would be in favor of moving the remainder to OSF. If we're going to use a lot of files in documentation then we could consider making some user-facing "fetcher" functions (e.g., `fetch_tutorial_data()`). If it's all just for testing then we don't have to worry as much, though!

### 2. Using `pytest.fixture`

@vinferrer took the first steps in using `pytest.fixture` functionality to consolidate downloading the files in `test_txt.py`, but I think we can use this to our advantage on a wider scale in the testing suite. `pytest` allows a `conftest.py` file in which you can put test-wide fixtures. I think we could move all the test downloads to that file and then they would be accessible to all the `test_X.py` modules. E.g.,

```python
@pytest.fixture(scope='session')
def multifreq_acq_file():
    url = 'https://osf.io/9a7yv/download'
    test_path = resource_filename('phys2bids', 'tests/data')
    test_filename = 'Test_belt_pulse_multifreq.acq'
    test_full_path = os.path.join(test_path, test_filename)
    wget.download(url, test_full_path)
    return test_full_path
```

Since the download steps are identical for each file I would recommend we make a small helper function, too:

```python
def fetch_file(url, filename):
    url = 'https://osf.io/{}/download'.format(url)
    test_path = resource_filename('phys2bids', 'tests/data')
    full_path = os.path.join(test_path, filename)
    if not os.path.isfile(full_path):
        wget.download(url, full_path)
    return full_path


@pytest.fixture(scope='session')
def multifreq_acq_file():
    return fetch_file('9a7yv', 'Test_belt_pulse_multifreq.acq')
```

Then we can have module-specific `pytest.fixture` functions that e.g., load these files (as @vinferrer did with `test_read_header_and_channels()` in `test_txt.py`.

(I would also suggest we download files to a temporary directory rather than to the `phys2bids/tests/data` directory, since that ends up cluttering the local git repo. We could use `pytest`'s [`tmp_path`](https://docs.pytest.org/en/latest/tmpdir.html) functionality to do this.)

### 3. Using `pytest.mark.parametrize`

There are a lot of times where we're using the same file and just testing slightly different parameters (e.g., changing the units in the header of a file and checking that the output of some function changes in response). This is the perfect use case for `pytest.mark.parametrize`! Having looked at `test_txt.py` most recently I know there are some functions where this could be used, but taking a quick peek at the test suite more broadly I think there are other places that would benefit from this, as well. E.g., looking at [`test_txt.test_process_acq()`](https://github.com/physiopy/phys2bids/blob/master/phys2bids/tests/test_txt.py#L105), I think something like this could work:

```python
@pytest.mark.parametrize('units, expected', [
    ('1 msec/sample', 1000),
    ('0.01 sec/sample', 100),
    ('1 µsec/sample', 1000000.0),
    ('100 Hz', 100),
    ('1 kHz', 1000),
    ('1 MHz', 1000000)
])
def test_process_acq(test_read_header_and_channels, units, expected):
    header = test_read_header_and_channels[0]
    channels = test_read_header_and_channels[1]
    chtrig = 2

    # set units to test that expected frequency is generated correctly
    header[1][0] = units
    phys_obj = txt.process_acq(channels, chtrig=chtrig, header=header)
    assert math.isclose(phys_obj.freq[0], expected)
```

## Next Steps
* Clean up `phys2bids/tests/data` directory to remove superfluous files
* Consolidate test file downloads into one module in the testing suite and make them accessible to other tests via `pytest.fixture` decorators
* Use `pytest.mark.parametrize` more liberally throughout testing suite

---

Sorry this got so long. Nonetheless, let me know your thoughts on all this! Looking forward to discussing with everyone :sparkles:

rmarkello:
Closes #179 

This PR makes (most) of the changes I described in #179 to demonstrate how it would change the test suite. It's a rather large PR (very sorry), largely because this is a pretty dramatic overhaul to the way testing had been set up. It's my hope that moving forward from this would reduce boilerplate code for tests (e.g., fetching new test data files) and streamline things.

## Proposed Changes

  - Removes `wget` dependency and consolidates fetching of all test data to [`phys2bids/tests/conftest.py`](https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions)
  - Modifies all test data downloads to use `pytest`'s temporary directory rather than going to `phys2bids/tests/data` to avoid modifying local git repos when testing locally
  - Updates all tests to use new test data fixtures
  - Adds use of `pytest.mark.parametrize` to some tests, as applicable 
  - Removes the two data files included in `phys2bids/tests/data` that were not actively used by any tests (the tutorial files are kept)

### DIFF
--- a/phys2bids/interfaces/acq.py
+++ b/phys2bids/interfaces/acq.py
@@ -5,6 +5,7 @@
 phys2bids interface for acqknowledge files.
 """
 import logging
+import warnings
 
 from bioread import read_file
 
@@ -33,7 +34,9 @@ def populate_phys_input(filename, chtrig):
     physio_obj.BlueprintInput
     """
 
-    data = read_file(filename).channels
+    with warnings.catch_warnings():
+        warnings.filterwarnings('ignore', category=DeprecationWarning)
+        data = read_file(filename).channels
 
     freq = [data[chtrig].samples_per_second, ]
     timeseries = [data[chtrig].time_index, ]
--- a/phys2bids/tests/conftest.py
+++ b/phys2bids/tests/conftest.py
@@ -4,8 +4,28 @@ from urllib.request import urlretrieve
 import pytest
 
 
-def fetch_file(url, path, filename):
-    url = 'https://osf.io/{}/download'.format(url)
+def fetch_file(osf_id, path, filename):
+    """
+    Fetches file located on OSF and downloads to `path`/`filename`1
+
+    Parameters
+    ----------
+    osf_id : str
+        Unique OSF ID for file to be downloaded. Will be inserted into relevant
+        location in URL: https://osf.io/{osf_id}/download
+    path : str
+        Path to which `filename` should be downloaded. Ideally a temporary
+        directory
+    filename : str
+        Name of file to be downloaded (does not necessarily have to match name
+        of file on OSF)
+
+    Returns
+    -------
+    full_path : str
+        Full path to downloaded `filename`
+    """
+    url = 'https://osf.io/{}/download'.format(osf_id)
     full_path = os.path.join(path, filename)
     if not os.path.isfile(full_path):
         urlretrieve(url, full_path)
--- a/phys2bids/tests/conftest.py
+++ b/phys2bids/tests/conftest.py
@@ -1,0 +1,60 @@
+import os
+from urllib.request import urlretrieve
+
+import pytest
+
+
+def fetch_file(url, path, filename):
+    url = 'https://osf.io/{}/download'.format(url)
+    full_path = os.path.join(path, filename)
+    if not os.path.isfile(full_path):
+        urlretrieve(url, full_path)
+    return full_path
+
+
+@pytest.fixture(scope='session')
+def testpath(tmp_path_factory):
+    """ Test path that will be used to download all files """
+    return tmp_path_factory.getbasetemp()
+
+
+@pytest.fixture
+def samefreq_full_acq_file(testpath):
+    return fetch_file('27gqb', testpath,
+                      'Test_belt_pulse_samefreq.acq')
+
+
+@pytest.fixture
+def samefreq_full_txt_file(testpath):
+    return fetch_file('sdz4n', testpath,
+                      'Test_belt_pulse_samefreq.txt')
+
+
+@pytest.fixture
+def samefreq_short_txt_file(testpath):
+    return fetch_file('4yudk', testpath,
+                      'Test_belt_pulse_samefreq_short.txt')
+
+
+@pytest.fixture
+def samefreq_noheader_txt_file(testpath):
+    return fetch_file('sre3h', testpath,
+                      'Test_belt_pulse_samefreq_no_header.txt')
+
+
+@pytest.fixture
+def multifreq_acq_file(testpath):
+    return fetch_file('9a7yv', testpath,
+                      'Test_belt_pulse_multifreq.acq')
+
+
+@pytest.fixture
+def multifreq_lab_file(testpath):
+    return fetch_file('q4x2f', testpath,
+                      'Test_2minRest_trig_multifreq_header_comment.txt')
+
+
+@pytest.fixture
+def notime_lab_file(testpath):
+    return fetch_file('u5dq8', testpath,
+                      'Test2_trigger_CO2_O2_pulse_1000Hz_534TRs_no_time.txt')
--- a/phys2bids/tests/test_integration.py
+++ b/phys2bids/tests/test_integration.py
@@ -1,12 +1,10 @@
-import os
-import wget
 import glob
 import json
 import math
+import os
+from pkg_resources import resource_filename
 import re
 import subprocess
-
-from pkg_resources import resource_filename
 
 from phys2bids.phys2bids import phys2bids
 from phys2bids._version import get_versions
@@ -102,16 +100,12 @@ def test_integration_tutorial():
         os.remove(filename)
 
 
-def test_integration_acq():
+def test_integration_acq(samefreq_full_acq_file):
     """
     Does the integration test for an acq file
     """
 
-    url = 'https://osf.io/27gqb/download'
-    test_path = resource_filename('phys2bids', 'tests/data')
-    test_filename = 'Test_belt_pulse_samefreq.acq'
-    test_full_path = os.path.join(test_path, test_filename)
-    wget.download(url, test_full_path)
+    test_path, test_filename = os.path.split(samefreq_full_acq_file)
     test_chtrig = 3
 
     phys2bids(filename=test_filename, indir=test_path, outdir=test_path,
@@ -153,16 +147,12 @@ def test_integration_acq():
         os.remove(filename)
 
 
-def test_integration_multifreq():
+def test_integration_multifreq(multifreq_acq_file):
     """
     Does the integration test for a multi-frequency file
     """
 
-    url = 'https://osf.io/9a7yv/download'
-    test_path = resource_filename('phys2bids', 'tests/data')
-    test_filename = 'Test_belt_pulse_multifreq.acq'
-    test_full_path = os.path.join(test_path, test_filename)
-    wget.download(url, test_full_path)
+    test_path, test_filename = os.path.split(multifreq_acq_file)
     test_chtrig = 3
 
     phys2bids(filename=test_filename, indir=test_path, outdir=test_path,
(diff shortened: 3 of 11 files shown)


rmarkello:
(Okay it looks like I deleted a TON of lines because I removed the test data files from `phys2bids/tests/data` 😱 Realistically it's...not _that_ much !)

smoia:
Since you are already working on tests, I'd like to ask you a favour:
Could you change the files used to test `viz` to [`Test_belt_pulse_samefreq.acq`](https://osf.io/27gqb/download)?
It's the same as the one used now (@vinferrer please confirm), but in `acq` format rather than in `txt`, and it weights a fourth of the one used now. It would speed up things a little!

eurunuela:
Wow, this looks great @rmarkello! May I suggest that @vinferrer and I review this PR?

vinferrer:
Sure I can review this, as soon as is marked as ready

phys2bids/tests/test_txt.py
```diff
@@ -1,185 +1,127 @@
...
     assert len(header_acq) == 8  # check proper header lenght
     assert len(channels_acq) == 1048560  # check proper number of timepoints
     assert len(header_acq[-1]) == 2  # check extra line is deleted
-    # load file with comment
-    # url to Test_2minRest_trig_multifreq_header_comment.txt
-    assert len(channels_lab[152109 - 9]) == 6  # check the comment has been eliminated
-    return header_acq, channels_acq, header_lab, channels_lab
+    assert 'acq' in header_acq[0][0]
 
+    return header_acq, channels_acq, chtrig
```
vinferrer: Hi Ross, the ``Test_2minRest_trig_multifreq_header_comment.txt`` was been used here because it has a comment iniside that the program has to eliminate with an if condition. Probably this is activated somewhere else now that you have refractored ```test_txt.py`` but are we sure?

vinferrer: I checked codecov and the line is tested so don't worry

phys2bids/tests/test_txt.py
```diff
@@ -1,185 +1,127 @@
...
+
+
+@pytest.mark.parametrize('units, expected', [
+    ('1 msec/sample', 1000),
+    ('0.01 sec/sample', 100),
+    ('1 µsec/sample', 1000000.0),
+    ('100 Hz', 100),
+    ('1 kHz', 1000),
+    ('1 MHz', 1000000)
+])
```
vinferrer: Could we add a test for ``min`` units? It's not being tested according to the codecov report and is one of the units accepted in the acq format

phys2bids/viz.py
```diff
@@ -66,7 +66,7 @@ def ntr2time(x):
     subplot.set_ylabel('Volts')
     subplot.plot(time, trigger, '-', time, thrline, 'r-.', time, block, '-')
     subplot.fill_between(time, block, where=block >= d, interpolate=True, color='#ffbb6e')
-    subplot.legend(["trigger", "input threshold", "time block"])
+    subplot.legend(["trigger", "input threshold", "time block"], loc='upper right')
```
vinferrer: Why are you changing this?

rmarkello: By default, `matplotlib.legend` defaults to `loc='best'`, where it attempts to place your legend in the best location based on the plotted data. When you have a _lot_ of data (as we often do in these plots) `matplotlib` will [raise a warning](https://github.com/matplotlib/matplotlib/blob/537a8c905e82fc24cb6d83042b30787aa396cd27/lib/matplotlib/legend.py#L1029-L1032) that using `loc='best'` is slower than manually specifying the location. Since (I think&mdash;but please correct me if I'm wrong!) the legend is supposed to always be in the upper right corner of the plot, setting the default location here should speed things up a bit.

phys2bids/tests/test_txt.py
rmarkello: For sure! Just something like `'0.001 min/sample'`? The expected number should be `16.6666666` or something like that, I believe, in that case. (I'll test it out to make sure.)

Let me know if you'd prefer something else.

phys2bids/viz.py
rmarkello: Let me know if you'd prefer that go in a separate PR! I was just doing it here because I was getting rid of all the warnings that the codebase was throwing in tests (e.g., I'm filtering out the `bioread` warning, as well).

rmarkello:
I think I'll hold off on making the tutorial data fetcher for another PR! This one seems big enough in scope as-is 😱 

But I think it's ready for review @eurunuela @vinferrer ! Thanks so much in advance 😄

phys2bids/interfaces/acq.py
```diff
@@ -33,7 +34,9 @@ def populate_phys_input(filename, chtrig):
     physio_obj.BlueprintInput
     """
 
-    data = read_file(filename).channels
+    with warnings.catch_warnings():
+        warnings.filterwarnings('ignore', category=DeprecationWarning)
```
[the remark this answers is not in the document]
rmarkello: This isn't the _best_ thing to do, but it was REALLY frustrating trying to read the test logs when we had >200 deprecation warnings :joy:

If I have time I'll try and make a PR upstream to `bioread`...

phys2bids/tests/test_integration.py
```diff
@@ -1,13 +1,11 @@
-import os
-import wget
 import glob
 import json
 import math
+import os
+from pkg_resources import resource_filename
 import re
 import subprocess
```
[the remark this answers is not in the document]
rmarkello: Yes, sorry! I reordered this without thinking... My default is:

1. Standard library imports
2. Third-party library imports
3. Internal library imports

Where each set of imports is in alphabetical order, ideally separated by a blank line 🤷‍♂

phys2bids/tests/conftest.py
```diff
@@ -0,0 +1,60 @@
+import os
+from urllib.request import urlretrieve
+
+import pytest
+
+
+def fetch_file(url, path, filename):
```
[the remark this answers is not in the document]
rmarkello: For sure!! I'll add a doc-string and maybe change the parameter name from `url` to `osf_id` or something to be a bit more clear.

phys2bids/interfaces/acq.py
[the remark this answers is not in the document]
eurunuela: Yeah... that's why I didn't do it in the first place but as you said it was really frustrating to read the logs haha.

phys2bids/tests/conftest.py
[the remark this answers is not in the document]
eurunuela: `osf_id` sounds perfect. Thanks!

phys2bids/tests/test_integration.py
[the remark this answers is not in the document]
eurunuela: I'll open an issue to discuss this.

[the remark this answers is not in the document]
eurunuela: #182 open to discuss this.

phys2bids/viz.py
vinferrer: It's fine, just asking

phys2bids/tests/test_txt.py
vinferrer: perfect

phys2bids/interfaces/acq.py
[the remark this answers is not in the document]
vinferrer: I agree, let's do it like this for the moment

vinferrer:
I think, as soon as you add the doc-string for ```fech_file`` this should be ready to go, nice work

Issue #173, opened 2020-03-12
New way of automatically selecting a threshold for the estimation of the number of time points


eurunuela:
## Detailed Description

The current part of the code that automatically selects and suggest a threshold for estimating the number of time points does not work properly. It calculates the first derivative of the trigger channel and sets a threshold on the positive values.

## Context / Motivation

Suggesting and automatically selecting a threshold would be a very nice and useful feature for the project.

## Possible Implementation

After a brief meeting yesterday, we thought of the following options to make this work:
1. Apply a threshold that is 2 times de STD of the trigger channel. This would be a very easy way of selecting the threshold, as 2 times de STD would get rid of the "noisy baseline" of the trigger channel.
2. The values in the trigger channel could be represented as two Gaussian distributions that ideally would not overlap. The threshold would be given by the point that separates the two distributions. This is a well-known problem in statistics and it is probably implemented in Python already. In my opinion, this would be the most elegant way of selecting the threshold.
3. Similar to the first option, we could base the threshold on the Median Absolute Deviation (MAD); i.e. the threshold would be set by the median value of the trigger channel plus the MAD.  I expect this option to be similar to the STD one in terms of functionality.
4. If we'd like to work with the first derivative, we should find the starting and ending point of the triggers (and not only the starting one, i.e. the positive one). If we have both the starting and ending points in the first derivative space and apply a threshold that removes the rest values, we could go back to the original space by integrating. This integration would be multiplying the trigger channel by a lower-triangular matrix of ones. Once in the original space, which would be clear of noise, we could select a threshold somewhere between the baseline (0) and the trigger amplitude.

Feel free to suggest possible solutions!

smoia:
The current ``master`` doesn't contain any code that automatically selects/suggests a threshold to detect the number of timepoitns. That code is contained in PR #153 .
However, we could add automatic detection of the number of timepoints or at least suggest a threshold to be used by the user. That would be a very nice feature to add to our code. In this case, we need to change the detection of the peaks as it's done now in ``physio_obj.py``.
As we discussed with @eurunuela and @vinferrer yesterday, I think as well that the option number two (leveraging the fact that our trigger channel is a binomial distribution) is the most elegant, and albeit computationally more intense, also most precise.
Alternatively, we need to use some statistics that is sensitive to outliers (the triggers should be outliers), hence a solution based on mean/standard deviation would be adequate.

rmarkello:
This is a great idea! 🙌 

[...]

PR #183, opened 2020-03-23
Adds check_trigger_amount_std

vinferrer:
check_tigger_amount_std detects the amount of triggers using the std

## Proposed Changes
  - This new function detects the trigger amount without the need of a threshold
  - It uses twice the std + the mean to generate a threshold
  - Then counts the blocks higher than that threshold which will be the amount of triggers
 -  it also corrects the timeoffset if necessary.
 -  One solution to issue #173


vinferrer:
Even if it says is ready for review, is not. We need to discuss the what we do with the other ``check_trigger_amount```function and then i still need to do the test

eurunuela:
Is the `check_trigger_amount` function really needed with the new approach? If it's not, I'd get rid of it and make the new one work for all use cases.

vinferrer:
So it depents on how good the new function works, that's why i said to keep it temporarily

vinferrer:
@smoia, @rmarkello what do you think about it?

eurunuela:
Well, then let's keep it and see how good the new function is! We will decide once we know and before we merge this PR.

vinferrer:
it's probably better than the last one

eurunuela:
If it's better than the previous one, we should only have the good one :)

smoia:
IMO I would keep both - I love the fact that we have an automatic way to get the threshold and guess the amount of timepoints and I think it should be the **default** running option.
However, I would also keep the possibility to override the default option and have a specified threshold in case something goes wrong.

From what I see in the code, the two functions are actually very similar - three quarter of the lines are actually _the same_. I would suggest to merge the two functions together, moving the "if one method else the other" statement inside (insert the new lines into the original function in a set of if .. else statements), and work in order to harmonise it with the rest of the code. I can help you with that if you want.

smoia:
We can even change the use of the derivative in the original function and use only the new point selection method. I like the fact that in this way, the threshold is related to original data.

phys2bids/phys2bids.py
```diff
@@ -176,7 +176,7 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
 
 def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
               sub=None, ses=None, chtrig=0, chsel=None, num_timepoints_expected=0,
-              tr=1, thr=2.5, ch_name=[], chplot='', debug=False, quiet=False):
+              tr=1, thr=2.5, ch_name=[], chplot='', debug=False, quiet=False, std_option=0):
```
rmarkello:
```suggestion
              tr=1, thr=2.5, ch_name=[], chplot='', debug=False, quiet=False, std_option=False):
```

I know that `0` and `1` are identical to `False` and `True` for all intents and purposes in Python, but since you currently have this set up as a binary option I think it would be better to use `std_option=False`.

phys2bids/phys2bids.py
```diff
@@ -261,7 +261,11 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
 
     # Run analysis on trigger channel to get first timepoint and the time offset.
     # #!# Get option of no trigger! (which is wrong practice or Respiract)
-    phys_in.check_trigger_amount(chtrig, thr, num_timepoints_expected, tr)
+    if std_option == 1:
```
rmarkello:
```suggestion
    if std_option:
```

Boolean check preferred over integer equality!

vinferrer:
Sure I can do that.

phys2bids/physio_obj.py
```diff
@@ -325,6 +325,79 @@ def print_info(self, filename):
 
         LGR.info(info)
 
+    def check_trigger_amount_std(self, chtrig=1, num_timepoints_expected=0, tr=0):
+        from itertools import groupby
```
smoia: I would move this import to the beginning of the file - itertools is a core module, so we know it will be installed, without adding specific dependencies, when phys2bids is installed!

phys2bids/physio_obj.py
```diff
@@ -325,6 +325,79 @@ def print_info(self, filename):
...
+            the time of first trigger.
+        """
+        LGR.info('Counting trigger points according to std method')
+        # Use twice the std of the trigger channel to get the threshold
+        # comparing it to a given threshold.
+        trigger = self.timeseries[chtrig]
+        thr_std = np.mean(trigger) + 2 * np.std(trigger)
+        timepoints = trigger > thr_std
+        num_timepoints_found = len([is_true for is_true, _ in groupby(timepoints,
+                                    lambda x: x != 0) if is_true])
```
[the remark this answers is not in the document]
smoia: @rmarkello true, but the `groupby` is necessary because triggers could be several timepoints long, and in that case you want to count it once only. Do you think that
` num_timepoints_found = np.sum(groupby(timepoints, lambda x: x != 0))`
would work?

phys2bids/physio_obj.py
vinferrer: Sure, I was putting it there for the moment because I tought it wasn't being used by any other function

phys2bids/phys2bids.py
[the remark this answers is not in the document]
vinferrer: I think in this case instead of adding an option we could activate the the previous function only if a threshold is given

phys2bids/physio_obj.py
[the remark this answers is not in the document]
rmarkello: Ahhh, totally missed that _very_ important detail!! No, `np.sum` won't work in that case. The solution @vinferrer came up with works! Ignore this suggestion.

phys2bids/phys2bids.py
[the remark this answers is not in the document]
vinferrer: So delete the std_option and put an action in the ```-thr`` option that activates the if. What do you think?

vinferrer:
I implemented the changes and mixed the functions

vinferrer:
I was thinking that probably is best for the user that if he defines the threshold instead of using the derivative, we directly use that threshold in the trigger channel and follow my method. This way there is no duplicity and everything is more straightforward

phys2bids/cli/run.py
```diff
@@ -110,8 +110,9 @@ def _get_parser():
                           dest='thr',
                           type=float,
                           help='Threshold to use for trigger detection. '
-                               'Default is 2.5.',
-                          default=2.5)
+                               'by default the rigger function calculates the needed threshold'
```
eurunuela:
```suggestion
                               'By default the trigger function calculates the needed threshold'
```

vinferrer:
In any case this PR should be accompanied with a change in the docs

phys2bids/physio_obj.py
```diff
@@ -261,9 +264,18 @@ def check_trigger_amount(self, chtrig=1, thr=2.5, num_timepoints_expected=0, tr=
...
+            trigger_deriv = np.diff(self.timeseries[chtrig])
+            timepoints = trigger_deriv > thr
+            num_timepoints_found = timepoints.sum()
+        else:
+            trigger = self.timeseries[chtrig]
+            thr = np.mean(trigger) + 2 * np.std(trigger)
+            timepoints = trigger > thr
+            num_timepoints_found = len([is_true for is_true, _ in groupby(timepoints,
+                                        lambda x: x != 0) if is_true])
+            LGR.info(f'number of expected timepoints according to std method'
```
eurunuela:
```suggestion
            LGR.info(f'The number of expected timepoints according to the std method'
```

phys2bids/physio_obj.py
```diff
@@ -261,9 +264,18 @@ def check_trigger_amount(self, chtrig=1, thr=2.5, num_timepoints_expected=0, tr=
...
+                     f'is {num_timepoints_found} the threshold is {thr}')
```
eurunuela:
```suggestion
                     f'is {num_timepoints_found} and the threshold is {thr}')
```

eurunuela:
Don’t worry about those tests. I’ll work on making them work right on this PR before we merge.

phys2bids/physio_obj.py
```diff
@@ -325,6 +337,79 @@ def print_info(self, filename):
 
         LGR.info(info)
 
+    def check_trigger_amount_std(self, chtrig=1, num_timepoints_expected=0, tr=0):
```
rmarkello: Should this function be deleted now, since you combined its functionality into `check_trigger_amount()`?

vinferrer: yeah i left it while testing in case i needed it

phys2bids/phys2bids.py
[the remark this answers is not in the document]
smoia: Mmm, I don't know. On one side, the user might want to run the program without automatic trigger count. On the other, we might have too many arguments, so deleting one is not a bad idea.    @rmarkello @eurunuela what do you think?

phys2bids/physio_obj.py
```diff
@@ -261,9 +264,18 @@ def check_trigger_amount(self, chtrig=1, thr=2.5, num_timepoints_expected=0, tr=
         LGR.info('Counting trigger points')
         # Use first derivative of the trigger channel to find the TRs,
         # comparing it to a given threshold.
-        trigger_deriv = np.diff(self.timeseries[chtrig])
-        timepoints = trigger_deriv > thr
-        num_timepoints_found = timepoints.sum()
+        if thr is not None:
```
smoia: I agree with you @vinferrer, let's see what happens if we completely drop the first derivative approach!
(you can drop this "if" statement)

phys2bids/phys2bids.py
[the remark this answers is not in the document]
eurunuela: We could make `thr=None` and run the automatic threshold in this case. If the user provides a threshold value, we don't run the automatic thing.

phys2bids/phys2bids.py
```diff
@@ -262,6 +262,7 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     # Run analysis on trigger channel to get first timepoint and the time offset.
     # #!# Get option of no trigger! (which is wrong practice or Respiract)
     phys_in.check_trigger_amount(chtrig, thr, num_timepoints_expected, tr)
+    thr = phys_in.thr
```
smoia: I would delete this line and use straight away `phys_in.thr` where needed (if I saw it right, it's just for the trigger plot, ~line 278)

phys2bids/phys2bids.py
```diff
@@ -262,6 +262,7 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     # Run analysis on trigger channel to get first timepoint and the time offset.
     # #!# Get option of no trigger! (which is wrong practice or Respiract)
     phys_in.check_trigger_amount(chtrig, thr, num_timepoints_expected, tr)
```
smoia: What if we move this line inside the if statement that follows?
In this way, if the users don't want to run the trigger count, they can _not_ specify the TR or the ntp and it won't take place.

phys2bids/physio_obj.py
```diff
@@ -296,7 +308,7 @@ def check_trigger_amount(self, chtrig=1, thr=2.5, num_timepoints_expected=0, tr=
         else:
             LGR.warning('The necessary options to find the amount of timepoints '
                         'were not provided.')
-
+        self.thr = thr
```
smoia: Can you add this Attribute in the docstring of the class?

phys2bids/cli/run.py
```diff
@@ -110,8 +110,9 @@ def _get_parser():
                           dest='thr',
                           type=float,
                           help='Threshold to use for trigger detection. '
-                               'Default is 2.5.',
-                          default=2.5)
+                               'By default the trigger function calculates the needed threshold'
```
smoia:
```suggestion
                               'If "ntp" and "TR" are specified, phys2bids automatically computes a threshold '
                               'to detect the triggers. Use this parameter to set it manually'
```

phys2bids/cli/run.py
```diff
@@ -110,8 +110,9 @@ def _get_parser():
...
+                               'use this parameter only if the default threshold fails',
```
smoia:
```suggestion

```

phys2bids/phys2bids.py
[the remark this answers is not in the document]
smoia: I was looking at the code in the review. If we move the trigger detection step inside an `if` statement that checks that the user provided `ntp` and `tr`, we should be ok.
In this way, if the users want any trigger detection (manual or not), they can add the other two parameters, otherwise drop it.

phys2bids/physio_obj.py
vinferrer: My approach is actually different, so if no thr is given then we do the std thing, if it is given then then we input that thr directly to the function. Therefore we always use the trigger instead of the derivative for the triggers amont detection

phys2bids/phys2bids.py
vinferrer: I disagree, if the trigger is not inputed then we use the thr from the function, with this line we avoid using an if statement in case the user inputs a trigger

phys2bids/physio_obj.py
vinferrer: This is already added to the docstring

phys2bids/physio_obj.py
```diff
@@ -250,6 +251,8 @@ def check_trigger_amount(self, chtrig=1, thr=2.5, num_timepoints_expected=0, tr=
         Notes
         -----
         Outcome:
+         self.thr:
+            Threshold used by the function to detect trigger points.
```
vinferrer: self.thr already in the docstring

phys2bids/phys2bids.py
vinferrer: I see your point okay we can do this

phys2bids/physio_obj.py
vinferrer: implemented

vinferrer:
@rmarkello something weird is happening. Here travis is not working for the newest commit, however if I go to the branch in my repo travis is executing okay. Why is this happening?

rmarkello:
Very unclear! I just re-triggered the build on Travis and it seems to be working okay now. Must have just been a stochastic error 🤷‍♂

eurunuela:
Thanks @rmarkello! @vinferrer let me know when you're finished so I can add the corrected tests to this PR.

vinferrer:
> 
> 
> Thanks @rmarkello! @vinferrer let me know when you're finished so I can add the corrected tests to this PR.
>

Well I am waiting for the last modifications to be approved, after that we will modify the docs, in that phase you should be able to start with the test mods since it's they are not going to be affected. I will tell you when the moment arrives.

phys2bids/physio_obj.py
smoia: Great! Can you add it to the docstring of the _class_ now?
It should go around [this line](https://github.com/physiopy/phys2bids/blob/386f3c2c51bf8774cb3dc91ad077290c90fa058d/phys2bids/physio_obj.py#L108), after the `num_timepoints_found` Attribute, that is computed in the same function.

phys2bids/physio_obj.py
```diff
@@ -250,6 +251,8 @@ def check_trigger_amount(self, chtrig=1, thr=2.5, num_timepoints_expected=0, tr=
         Notes
         -----
         Outcome:
+         self.thr:
```
smoia:
```suggestion
        self.thr:
```

phys2bids/phys2bids.py
```diff
@@ -259,13 +259,13 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
...
-    # #!# Get option of no trigger! (which is wrong practice or Respiract)
-    phys_in.check_trigger_amount(chtrig, thr, num_timepoints_expected, tr)
-
     # Create trigger plot. If possible, to have multiple outputs in the same
     # place, adds sub and ses label.
     if tr != 0 or num_timepoints_expected != 0:
+        # Run analysis on trigger channel to get first timepoint and the time offset.
+        # #!# Get option of no trigger! (which is wrong practice or Respiract)
+        phys_in.check_trigger_amount(chtrig, thr, num_timepoints_expected, tr)
+        thr = phys_in.thr
```
smoia: I reply in this review because unluckily I can't reply in the general conversation.
I see what you mean, but the only other place `thr` is called now is [this line](https://github.com/physiopy/phys2bids/blob/386f3c2c51bf8774cb3dc91ad077290c90fa058d/phys2bids/phys2bids.py#L276) that is in the same `if` statement.
If I understand the program as it is now, there is no valide possible case in which `viz.plot_trigger` is not exectued without  `phys_in.check_trigger_amount(chtrig, thr, num_timepoints_expected, tr)` being executed first. `check_trigger_amount` creates the attribute that is specified here, `phys_in.thr`, no matter whether the user provides a threshold manually or not (the attribute gets created at the end of the function). So there is no valid scenario in which, when you call `viz.plot_trigger`, `phys_in.thr` doesn't exist.
It's not a major deal, so if you really want to keep this line, we can. I just don't see the reason to have it, that's all.

vinferrer: Point taken. Line eliminated

phys2bids/physio_obj.py
vinferrer: Done

vinferrer:
Okay, I think @eurunuela can start with the test_integration mod to make it work with the new function. I am going to modify the docs so the changes get reflected

phys2bids/phys2bids.py
```diff
@@ -259,13 +259,12 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
...
 
-    # Run analysis on trigger channel to get first timepoint and the time offset.
-    # #!# Get option of no trigger! (which is wrong practice or Respiract)
-    phys_in.check_trigger_amount(chtrig, thr, num_timepoints_expected, tr)
-
     # Create trigger plot. If possible, to have multiple outputs in the same
     # place, adds sub and ses label.
     if tr != 0 or num_timepoints_expected != 0:
+        # Run analysis on trigger channel to get first timepoint and the time offset.
+        # #!# Get option of no trigger! (which is wrong practice or Respiract)
```
vinferrer: I just detected a bug, if ``-tr`` is not inputed and ``-ntp`` is inputed this happens:
```
phys2bids -in tutorial_file.txt -chtrig 1 -outdir outdir -ntp 158 
INFO:phys2bids.phys2bids:Currently running phys2bids version v1.3.0-beta+335.gca4394e.dirty
INFO:phys2bids.phys2bids:Input file is tutorial_file.txt
INFO:phys2bids.utils:File extension is .txt
WARNING:phys2bids.utils:If both acq and txt files exist in the path, acq will be selected.
INFO:phys2bids.phys2bids:Reading the file ./tutorial_file.txt
INFO:phys2bids.interfaces.txt:phys2bids detected that your file is in Labchart format
INFO:phys2bids.phys2bids:Reading infos
INFO:phys2bids.physio_obj:
------------------------------------------------
File tutorial_file.txt contains:
01. Trigger; sampled at 1000.0 Hz
02. CO2; sampled at 1000.0 Hz
03. O2; sampled at 1000.0 Hz
04. Pulse; sampled at 1000.0 Hz
------------------------------------------------

INFO:phys2bids.physio_obj:Counting trigger points
INFO:phys2bids.physio_obj:The number of timepoints according to the std method is 158 and the threshold is 1.1523587407910223
INFO:phys2bids.physio_obj:Checking number of timepoints
INFO:phys2bids.physio_obj:Found just the right amount of timepoints!
INFO:phys2bids.phys2bids:Plot trigger
/home/vicente/phys2bids/phys2bids/viz.py:72: UserWarning: Attempting to set identical left == right == 0 results in singular transformations; automatically expanding.
  subplot.set_xlim([-tr * 4, tr * 4])
/home/vicente/phys2bids/phys2bids/viz.py:48: RuntimeWarning: invalid value encountered in true_divide
  return x / tr
/home/vicente/phys2bids/phys2bids/viz.py:48: RuntimeWarning: invalid value encountered in true_divide
  return x / tr
/home/vicente/phys2bids/phys2bids/viz.py:48: RuntimeWarning: divide by zero encountered in true_divide
  return x / tr
Traceback (most recent call last):
  File "/home/vicente/.local/bin/phys2bids", line 11, in <module>
    load_entry_point('phys2bids', 'console_scripts', 'phys2bids')()
  File "/home/vicente/phys2bids/phys2bids/phys2bids.py", line 362, in _main
    phys2bids(**vars(options))
  File "/home/vicente/phys2bids/phys2bids/phys2bids.py", line 276, in phys2bids
    plot_path, tr, phys_in.thr, num_timepoints_expected, filename)
  File "/home/vicente/phys2bids/phys2bids/viz.py", line 113, in plot_trigger
    plt.savefig(fileprefix + '_trigger_time.png', dpi=dpi)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/pyplot.py", line 729, in savefig
    res = fig.savefig(*args, **kwargs)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/figure.py", line 2180, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/backend_bases.py", line 2091, in print_figure
    **kwargs)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/backends/backend_gtk3agg.py", line 82, in print_png
    return agg.print_png(filename, *args, **kwargs)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/backends/backend_agg.py", line 527, in print_png
    FigureCanvasAgg.draw(self)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/backends/backend_agg.py", line 388, in draw
    self.figure.draw(self.renderer)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/artist.py", line 38, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/figure.py", line 1691, in draw
    child.apply_aspect(pos)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/axes/_secondary_axes.py", line 180, in apply_aspect
    self._set_lims()
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/axes/_secondary_axes.py", line 311, in _set_lims
    set_lim(lims)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/axes/_base.py", line 3241, in set_xlim
    left = self._validate_converted_limits(left, self.convert_xunits)
  File "/home/vicente/.local/lib/python3.6/site-packages/matplotlib/axes/_base.py", line 3158, in _validate_converted_limits
    raise ValueError("Axis limits cannot be NaN or Inf")
ValueError: Axis limits cannot be NaN or Inf
```
I believe the best way to avoid this behavior is that instead of a or condition, we have an and condition. We could also add an if so the viz.plot_trigger function does not get ``tr=None``

vinferrer: This is an error inherit from #153 we can deal with it in a bug PR

smoia: @vinferrer you're totally right. Can you implement the `and` statement instead of the `or` in this PR?

smoia:
@vinferrer, soon we're going to start using `auto` to automatically maintain versions, changelogs, and contributions. As the tutorial needs a bit more rewriting, could you please move your commit to change the tutorial to a new PR, labelled `Documentation`?
In that way if we need a bit more revision to integrate it in the text, we can work a bit longer and merge this in the meantime!
(also, I noticed the messages the function emits, and we might want to make it a bit clearer)

phys2bids/physio_obj.py
```diff
@@ -259,11 +262,16 @@ def check_trigger_amount(self, chtrig=1, thr=2.5, num_timepoints_expected=0, tr=
...
-        trigger_deriv = np.diff(self.timeseries[chtrig])
-        timepoints = trigger_deriv > thr
-        num_timepoints_found = timepoints.sum()
+        trigger = self.timeseries[chtrig]
+        if thr is None:
+            thr = np.mean(trigger) + 2 * np.std(trigger)
+        timepoints = trigger > thr
+        num_timepoints_found = len([is_true for is_true, _ in groupby(timepoints,
+                                    lambda x: x != 0) if is_true])
+        LGR.info(f'The number of expected timepoints according to the std method'
```
smoia:
```suggestion
        LGR.info(f'The number of expected timepoints according to the std_thr method'
```

phys2bids/physio_obj.py
```diff
@@ -259,11 +262,16 @@ def check_trigger_amount(self, chtrig=1, thr=2.5, num_timepoints_expected=0, tr=
...
+                 f'is {num_timepoints_found} and the threshold is {thr}')
```
smoia:
```suggestion
                 f'is {num_timepoints_found} and the computed threshold is {thr}')
```

vinferrer:
> @vinferrer, soon we're going to start using `auto` to automatically maintain versions, changelogs, and contributions. As the tutorial needs a bit more rewriting, could you please move your commit to change the tutorial to a new PR, labelled `Documentation`?
> In that way if we need a bit more revision to integrate it in the text, we can work a bit longer and merge this in the meantime!
> (also, I noticed the messages the function emits, and we might want to make it a bit clearer)

Commit undone

vinferrer:
We will open a new PR for the documentation

smoia:
Great!
Once you address all of @eurunuela's requested changes and mine, we will merge this PR in!

phys2bids/physio_obj.py
vinferrer: done

phys2bids/physio_obj.py
vinferrer: done

vinferrer:
I think they are adressed, what we need now is modify ``test_integration.py`` so travis works

vinferrer:
@eurunuela do you wanna take care of this?

eurunuela:
Yes, I'll work on it either today or tomorrow.

eurunuela:
Okay guys, there's a little bug. 

`phys_in.num_timepoints_found` only exists `if tr != 0 and num_timepoints_expected != 0`. However, it is always passed as an argument in the [print_summary call](https://github.com/physiopy/phys2bids/blob/9ec9c33da2e7c3adf7653a79cb024ffcc130bf42/phys2bids/phys2bids.py#L357).

So, there are two options that I can think of to fix this:

1. We initialize `num_timepoints_found = None` or `num_timepoints_found = 0`. 
2. We write an if statement that only passes this argument to the `print_summary()` function when `phys_in.num_timepoints_found` exists.

For the moment, I'll commit the fixes to the other two tests.

smoia:
Uff, good catch @eurunuela . I would opt for option number 2, printing a statement such as "No trigger count run" to the log and to the screen. We should also report the threshold in the same statement now that I think about it!.

eurunuela:
Shall we open another PR that takes care of that and merge this one? I think they're unrelated.

smoia:
@vinferrer @rmarkello your thoughts?

rmarkello:
I think I lean towards defaulting to have it set to `None` rather than an `if` statement, but am happy to defer to others! 

My reasoning: the `if` statement would have to handle passing `phys_in.num_timepoints_found` to `print_summary()` *and* `print_summary()` would have to be modified to change functionality when it doesn't receive that parameter. Setting `phys_in.num_timepoints_found` to `None` at the top of the class seems like a smaller change 🤷‍♂

vinferrer:
I think I agree with @rmarkello

smoia:
@eurunuela if you point out which test is not passing, then we can check that the others do and merge this in!

vinferrer:
> 
> 
> Shall we open another PR that takes care of that and merge this one? I think they're unrelated.

They are not unrelated it happened because I put the check_trigger_amount inside the if statement

vinferrer:
let's try by setting num_timepoints_found to None

eurunuela:
The logger test fails because of the bug I mentioned earlier.

vinferrer:
solved

smoia:
Thank you @vinferrer . Can you update the docstring accordingly?

smoia:
(or I can do it through a PR to your branch)

vinferrer:
where does that go?

vinferrer:
i may do it tomorrow

smoia:
Line 105 should be updated as `int or None`, and then the relative text that should be adapted.

vinferrer:
oh okay

phys2bids/physio_obj.py
```diff
@@ -102,7 +102,7 @@ class BlueprintInput():
         in the output files.
     units : (ch) list of strings
         List of the units of the channels.
-    num_timepoints_found: int
+    num_timepoints_found: int or None
         Amount of timepoints found in the automatic count.
         This is computed internally, *if* check_trigger_amount() is run
```
smoia:
```suggestion
        This is initialised as "None" and then computed internally,
        *if* check_trigger_amount() is run
```

eurunuela:
Awesome job, thanks guys! 🙌🏻

vinferrer:
Thank you for addressing the test problems

vinferrer:
let's merge it

PR #187, opened 2020-03-26
Quick fix for #184

### LINKED ISSUE #184: Output files are not BIDS compliant
## Expected Behavior

Following [BIDS 1.2.2](https://bids-specification.readthedocs.io/en/v1.2.2/), the `.tsv.gz` files should have a column for each channel.

## Actual Behavior

Our output has a row for each channel

## Possible solution

We can solve this in two ways:
1. a very easy very quick `BlueprintOutput.timeseries` transpose before (during) the final export to `tsv`, or
2. a less easy, less quick inversion of `BlueprintOutput.timeseries` indexes, followed by adaptation of the scripts, so that `BlueprintOutput.timeseries` will be always created and managed as if we were importing a BIDS compliant `tsv` file.

I prefer option number 2, that could be a bit less easy, but will work very well in case we need to import files or we need this object to be used in other libraries that only import BIDS compliant files. But I'd like to know what y'all think (starting with code contributors @vinferrer @rmarkello @eurunuela)

(thanks to @sangfrois for pointing it out)

smoia:
Closes #184 

## Proposed Changes

  - timeseries matrix gets transposed during its export, to be BIDS compliant.

(thanks to @sangfrois for pointing it out)

### DIFF
--- a/phys2bids/phys2bids.py
+++ b/phys2bids/phys2bids.py
@@ -348,7 +348,7 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
                 outfile = f'{outfile}_{uniq_freq}'
 
         LGR.info(f'Exporting files for freq {uniq_freq}')
-        savetxt(outfile + '.tsv.gz', transpose(phys_out[uniq_freq].timeseries),
+        savetxt(outfile + '.tsv.gz', phys_out[uniq_freq].timeseries.T,
                 fmt='%.8e', delimiter='\t')
         print_json(outfile, phys_out[uniq_freq].freq,
                    phys_out[uniq_freq].start_time,
--- a/phys2bids/phys2bids.py
+++ b/phys2bids/phys2bids.py
@@ -32,7 +32,7 @@ import logging
 import datetime
 
 from copy import deepcopy
-from numpy import savetxt
+from numpy import savetxt, transpose
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


phys2bids/phys2bids.py
```diff
@@ -348,7 +348,7 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
                 outfile = f'{outfile}_{uniq_freq}'
 
         LGR.info(f'Exporting files for freq {uniq_freq}')
-        savetxt(outfile + '.tsv.gz', phys_out[uniq_freq].timeseries,
+        savetxt(outfile + '.tsv.gz', transpose(phys_out[uniq_freq].timeseries),
```
rmarkello:
```suggestion
        savetxt(outfile + '.tsv.gz', phys_out[uniq_freq].timeseries.T,
```

phys2bids/phys2bids.py
```diff
@@ -32,7 +32,7 @@
 import datetime
 
 from copy import deepcopy
-from numpy import savetxt
+from numpy import savetxt, transpose
```
rmarkello:
```suggestion
from numpy import savetxt
```

smoia:
Sure thing!

smoia:
@rmarkello when you agree, you can merge it in!

PR #195, opened 2020-04-02
Reordering of import statements in the entire phys2bids project

### LINKED ISSUE #182: Discussion on the ordering of library imports
## Summary

Each one of us has got a different way of ordering the library imports and I believe we should have some guidelines to follow. The idea is that we are all on the same page. Also, this should make reading and reviewing the code easier on the eye.

The following is a possible guideline:
1. Standard library imports
2. Third-party library imports
3. Internal library imports

where each set of imports is in alphabetical order, ideally separated by a blank line.

If you agree on making the suggested guideline the default, please upvote. If you have a better suggestion, please comment on this issue. I will keep this issue open for a week and will open a PR that updates the imports with whatever guideline has got the most upvotes once the issue is closed.

Tagging the most active members in the project: @smoia, @vinferrer, @rmarkello, @RayStick.

## Next Steps

- Open a PR that updates the imports following the outcome of this discussion.

eurunuela:
Closes #182.

## Proposed Changes

  - Reorders imports as approved in #182.

### DIFF
--- a/phys2bids/interfaces/txt.py
+++ b/phys2bids/interfaces/txt.py
@@ -6,12 +6,13 @@ phys2bids interface for txt files.
 """
 
 import logging
-
 import numpy as np
 
-from phys2bids.physio_obj import BlueprintInput
-from operator import itemgetter
 from collections import Counter
+from operator import itemgetter
+
+from phys2bids.physio_obj import BlueprintInput
+
 
 LGR = logging.getLogger(__name__)
 
--- a/phys2bids/interfaces/txt.py
+++ b/phys2bids/interfaces/txt.py
@@ -6,10 +6,10 @@ phys2bids interface for txt files.
 """
 
 import logging
-import numpy as np
-
 from collections import Counter
 from operator import itemgetter
+
+import numpy as np
 
 from phys2bids.physio_obj import BlueprintInput
 
--- a/phys2bids/interfaces/txt.py
+++ b/phys2bids/interfaces/txt.py
@@ -13,7 +13,6 @@ import numpy as np
 
 from phys2bids.physio_obj import BlueprintInput
 
-
 LGR = logging.getLogger(__name__)
 
 
--- a/phys2bids/tests/conftest.py
+++ b/phys2bids/tests/conftest.py
@@ -1,7 +1,6 @@
 import os
-from urllib.request import urlretrieve
-
 import pytest
+from urllib.request import urlretrieve
 
 
 def fetch_file(osf_id, path, filename):
--- a/phys2bids/tests/conftest.py
+++ b/phys2bids/tests/conftest.py
@@ -1,6 +1,7 @@
 import os
-import pytest
 from urllib.request import urlretrieve
+
+import pytest
 
 
 def fetch_file(osf_id, path, filename):
--- a/phys2bids/tests/test_integration.py
+++ b/phys2bids/tests/test_integration.py
@@ -2,12 +2,13 @@ import glob
 import json
 import math
 import os
-from pkg_resources import resource_filename
 import re
 import subprocess
 
-from phys2bids.phys2bids import phys2bids
+from pkg_resources import resource_filename
+
 from phys2bids._version import get_versions
+from phys2bids.phys2bids import phys2bids
 
 
 def check_string(str_container, str_to_find, str_expected, is_num=True):
--- a/phys2bids/tests/test_integration.py
+++ b/phys2bids/tests/test_integration.py
@@ -4,7 +4,6 @@ import math
 import os
 import re
 import subprocess
-
 from pkg_resources import resource_filename
 
 from phys2bids._version import get_versions
--- a/phys2bids/tests/test_phys2bids.py
+++ b/phys2bids/tests/test_phys2bids.py
@@ -4,7 +4,6 @@ Tests phys2bids.py
 
 import json
 import os
-
 from pkg_resources import resource_filename
 
 from phys2bids import phys2bids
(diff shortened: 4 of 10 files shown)


smoia:
I know that I took myself out from the review task, but I had a peek this morning at it.
To my understanding, if we want to follow PEP8 conventions as you proposed in #182, the import should be a bit different. libraries such as `collections` and `operator` are standard, while `numpy` is third party.
For instance, taking the `txt.py` imports, they should follow:
```
# standard libraries
import logging
from collections import Counter
from operator import itemgetter

# third parties
import numpy as np

# internal
from phys2bids.physio_obj import BlueprintInput
```

It's quite tricky to categorise "standard libraries", as we use much more often packages like `numpy` and `matplotlib`, that are third parties, rather than `operator` or `collections` that are standard. I found a list [here](https://docs.python.org/3.6/library/). I think that another rule of thumb is that if it is in our requirement, then it's a third party. However, it's not easy to say for me either 
Maybe @rmarkello could review this PR?

eurunuela:
Good point! I always find it difficult to tell when it's a standard library or a third-party one. To me, they're all third-party 🤣

rmarkello:
Happy to review ! 

@smoia: that's the right list you linked to, and indeed&mdash;if it's in the requirements.txt file then it's third-party (assuming you didn't add any standard library modules to that list 🙈)!

@eurunuela: do you wanna do another re-ordering based on the list @smoia linked describing what modules are standard and then let me know when to take a look at this?

eurunuela:
Sure thing @rmarkello! I’ll let you know when it’s ready.

eurunuela:
@rmarkello forgot to tell you this last commit should've addressed the comments in our previous discussion

phys2bids/interfaces/txt.py
```diff
@@ -6,12 +6,13 @@
...
 import logging
+from collections import Counter
+from operator import itemgetter
 
 import numpy as np
 
 from phys2bids.physio_obj import BlueprintInput
-from operator import itemgetter
-from collections import Counter
+
```
rmarkello, line 15:
```suggestion
```

You don't need two blank lines here. You do before function statements, but not before the definition of e.g. "constants."

phys2bids/tests/conftest.py
```diff
@@ -1,7 +1,6 @@
 import os
-from urllib.request import urlretrieve
-
 import pytest
```
rmarkello, line 2: `pytest` is not part of the standard library and should be separate into a third-part imports section.

phys2bids/tests/test_integration.py
```diff
@@ -2,12 +2,13 @@
 import json
 import math
 import os
-from pkg_resources import resource_filename
 import re
 import subprocess
```
rmarkello, line 7:
```suggestion
```

`pkg_resources` _is_ part of the standard library and should be bundled with the standard imports. It mostly is here, but we just need to get rid of this extra blank line!

phys2bids/tests/test_phys2bids.py
```diff
@@ -2,11 +2,13 @@
 Tests phys2bids.py
 """
 
-import os
 import json
-from phys2bids import phys2bids
+import os
+
```
rmarkello, line 7:
```suggestion
```

Same as the last one; no blank line between the `import` and `from X import Y`

phys2bids/tests/test_physio_obj.py
```diff
@@ -2,9 +2,11 @@
 Tests physio_obj.py
 """
 
+from pytest import raises
```
rmarkello, line 5: Same as above, `pytest` is a third-part import: it should go below `numpy` here because you're doing a `from X import Y` statement.

phys2bids/tests/test_txt.py
```diff
@@ -1,5 +1,4 @@
 import math
```
rmarkello, line 1:
```suggestion
import math

```

Need the blank line; `pytest` is third-party.

eurunuela:
@rmarkello I'm sorry I made you "waste" your time like that. Apparently my last changes were not correctly pushed (I had just factory reset my computer so something probably did not go as expected with git).

Most of your comments had already been assessed, so this time it should be even easier to review.

Thank you so much! 💯

phys2bids/tests/conftest.py
rmarkello: This point still stands! `pytest` is not part of the stdlib and should be moved to a third-party library import section.

phys2bids/tests/test_integration.py
rmarkello: This also still stands! `pkg_resources` shouldn't be separated from standard library imports.

phys2bids/tests/test_phys2bids.py
rmarkello: Same as above! `pkg_resources` needs to be part of the standard library imports

phys2bids/tests/test_txt.py
rmarkello: Need this blank line. `pytest` is third-party!

rmarkello:
Okay, @eurunuela ! That review was a bit weird because I was commenting on issues that had been marked as "resolved" so it's showing them as...still resolved. I unresolved all the still-present issues from my first review—take a look at those (there should be ~4) and let me know what you think!

Thanks so much!

eurunuela:
Okay @rmarkello, I think we're both on the same page now. I'm sorry for the misunderstanding.

Thanks!

eurunuela:
Thanks @rmarkello ! I'll merge it myself. I don't think this issue needs another review (it's a minimal change).

PR #208, opened 2020-04-09
Add Azure Pipelines for Windows CI testing

### LINKED ISSUE #174: Add Windows CI testing

rmarkello:
Closes #174

## Proposed Changes

- Adds relevant Azure pipelines files to set up testing on Windows


rmarkello:
Alright, so because there's no Windows wheel for `bioread` on PyPi it's trying to install from the sdist, which is leading to this installation error. The only way around it is going to be to install `numpy` first and _then_ install `bioread`, which is a little annoying.

Another potential fix to upstream to `bioread`! :grin:

rmarkello:
Alright, I think this is ready to go for a review ! Let me know if you have any questions about anything :grin:

smoia:
This is great!
One thing: I noticed in azure that the [codecov upload](https://dev.azure.com/physiopy/phys2bids/_build/results?buildId=10&view=logs&j=2c855bde-037d-57e7-f7a3-61fc65066dac&t=dd71b88c-512f-5964-a9d7-f49e4ccf9e5a) seems to succeed but actually fails internally.
I don't think it's necessary to have it though: we already measure code coverage with Travis. Shouldn't it be enough?

.azure-pipelines/windows.yml
```diff
@@ -0,0 +1,48 @@
...
+        python -m pip install .[$(CHECK_TYPE)]
+      displayName: 'Install phys2bids'
+    - script: |
+        mkdir for_testing
+        cd for_testing
+        cp ../setup.cfg .
+        python -m pytest --doctest-modules --cov-report term-missing --cov=phys2bids --pyargs phys2bids
+      displayName: 'Run tests'
+      condition: and(succeeded(), eq(variables['CHECK_TYPE'], 'test'))
+    - script: |
```
smoia, line 42: Maybe I don't understand it, but I don't think we need this. Should it be different from the Travis codecov anyway?

smoia:
@rmarkello, I will also need you to delete the `[TEST]` from the title of the PR before merging!

.azure-pipelines/windows.yml
rmarkello: In theory it _shouldn't_ be different but there's a possibility that different OSs could hit different lines 🤷‍♂️ 

I'm not sure I see the harm in uploading it either way! Let me know if you have a reason that you wouldn't want it uploaded.

smoia: Well, for one I don't know how it would behave - will we have two codecov reports, or will we get the report from the pipeline that finishes last/first?
BUT, in case we want it, then I'm sorry to say that I guess you'll have to check it as it's not going through :sweat_smile:

rmarkello: Are you sure? Azure _thinks_ it's going through (see [relevant log here](https://dev.azure.com/physiopy/phys2bids/_build/results?buildId=10&view=logs&j=2c855bde-037d-57e7-f7a3-61fc65066dac&t=dd71b88c-512f-5964-a9d7-f49e4ccf9e5a&l=56))... And codecov generally [merges reports together](https://docs.codecov.io/docs/merging-reports) without any issues.

smoia: Yup, you're right, sorry!
Do you think it's feasible to:
1. Correct the errors that pop up in this [log](https://dev.azure.com/physiopy/phys2bids/_build/results?buildId=10&view=logs&j=2c855bde-037d-57e7-f7a3-61fc65066dac&t=dd71b88c-512f-5964-a9d7-f49e4ccf9e5a&l=56) and
2. Make the reports from Azure to clearly state they're from Azure, instead of being "Unknowns"?  
If not, we can merge this in and open issues about those two points for later work.

rmarkello:
This is failing now because one of the testing files was removed from OSF (https://osf.io/u5dq8/). @RayStick was that an accident or is there a file that should replace it? Happy to change which file is downloaded in the tests if so!

RayStick:
Oh, sorry! I removed some files that I thought were not being used by anyone. (I will make sure to check before removing/renaming next time). 

I replaced these files with a folder **all_exports** that has a much more bigger selection of exports of the same original file. So if you want a file with one frequency, header info and no time (as you did before), I would try [this one](https://osf.io/5829m/). 

If there are problems, let me know.

RayStick:
@rmarkello It would be better to use these new files, however if you need the other files back I can try to do that.

rmarkello:
We can use the new files, no issues! We just have to update the relevant info in the tests. I'll do that some time next Monday and let you know if I have any issues. Thanks for the quick response! 😄

RayStick:
@rmarkello I actually put these delete files back, anyway.
So you could use the files you were using before, to test (though the hyperlinks will need to be updates) or you can use files in 'all_exports' folder.

rmarkello:
@smoia this is the best codecov is gonna get. All the uploads are "Untitled" but it's better than nothing (and honestly, I don't think most people usually look at the "Builds" tab—usually you just look at the integrated reports).

If that works for you go ahead and merge :+1:

smoia:
That's alright @rmarkello, thank you so much for trying!
I'm going to close and merge!

rmarkello:
Wait why close? Is this not okay to be merged in?

smoia:
Sorry - definitely wrong button! I meant I'm going to _merge_ it in and that's all!

PR #206, opened 2020-04-07
Add the possibility to split multi-run physiological recordings

sangfrois:
Close #36
These changes are made so that phys2bids workflow can split multi-run recordings into different `BlueprintInput` objects based on run's start and end indexes (plus or minus a padding given by the user), then deal with each one of them independently - and save them in BIDS format.

### Adjust the CLI
-  in `run.py` script : change parser arguments.
    - [x] `-ntp` becomes a list of number of trigger timepoints expected in each run.
    - [x] `-tr` becomes a list of TR (maybe there are different sequences).
    - [x] `-padding` is an optional value that contains the padding to add before and after each run (default to 9s).
### Adjust Blueprint objects
- [x] The outputs of `check_trigger_amount` in BlueprintInput are now arguments that can be passed to the object, since they have to be overwritten if multi-runs are present.
- [x] `filename`, the name of the output file without extension, becomes an attribute of the BlueprintOutput object. 
### Adjust the main workflow
- in `phys2bids.py`, integrate a section dedicated to handle multi-run user inputs.
    - [x] Check `tr` and `ntp` list length equivalency and raise appropriate errors. If `tr` has one element, repeat it for the numbers of runs in `-ntp`.
    - [x] Check if the sum of expected timepoints equals `num_timepoints_found`, if not raise an error and stop (otherwise the risk of propagating errors is high).
    - [x] `phys_in` becomes a dictionary after the workflow checks whether there are multiple runs or not, to better handle the possibility of multiple runs.
    - [x] `phys_out` keys become based on frequencies and runs (to handle multi-frequency multi-run files).
    - [x] In order to avoid overwriting, the function checks if the name of an output is already in use between different stances of `phys_out`.
### Changes to `bids_unit.py` and creation of a function that deals with all the bids-related steps
- [x] Rename `bids_unit.py` to `bids.py`. **Equally, the tests change names too**
- [x] Move the handle of heuristic and renaming in this new function. **The tests for that function change place too!**
- [x] In order to avoid overwriting, 
### Change to heuristics
- [x] heuristics now can optionally accept `run` as an argument to deal with multiple runs.
### Adjust `viz.py`
- [x] There is now a function that calls `plot_trigger` (due to the repetition of code in the main workflow) that is called from the main workflow.
### Add a utility that finds runs start and end, split the input object into multiple input objects
- in `splice4phys.py` script
   - [x] A new function that find beginning and ends of runs `find_runs`
[...]

PR #209, opened 2020-04-10
Make heuristics a simple set of "if .. elif .. else" statements and use a dictionary instead of variables.

### LINKED ISSUE #44: Making heuristic functions a simple set of if elif else statements
## Detailed Description

At the moment, heuristic files contain a bunch of code that shouldn't be changed by the user (e.g. from [here](https://github.com/smoia/phys2bids/blob/f2d7d52e4e9d9d963f98aeb8809c81660e2f1794/phys2bids/heuristics/heur_test_acq.py#L30)).
It would be better to move that code outside of the heur file, so that the only content of those scripts is a simple definition that reads an input, goes through a bunch of `if` statements to set up the labels necessary for the name of the file, and returns those labels. 

## Context / Motivation

This would simplify the heuristic files on one hand, and help the integration of phys2bids in other BIDSificators programs such as BIDScoin (as in #199).

## Possible Implementation

- Move all the non-`if..elif..else` code into the function [`use_heuristics`](https://github.com/physiopy/phys2bids/blob/8420e44305f17719ff6c49b2f14deb59a6191897/phys2bids/phys2bids.py#L114).

### LINKED ISSUE #89: 0- vs 1-indexing the `run` keyword in `use_heuristic()`
In https://github.com/physiopy/phys2bids/pull/76#discussion_r356095934 a brief discussion was started about whether the `run` keyword should be 0- or 1-indexed when users provide a `heuristic` to convert their data to BIDS format. We are currently using a 0-indexing approach; I personally think a 1-indexing approach makes more sense here, but am open to discussion on this point!

@smoia found a relevant portion of the [BIDS spec](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#participant-names-and-other-labels) that touches on this, but the spec is not prescriptive on this point. I'm happy to hear others thoughts and go about issuing a change if it's decided that's warranted!

smoia:
Closes #89, #44 

## Proposed Changes

  - All the heuristics set run to  `run-01` rather than `run-00` (closes #89 )
  - The tests have been updated consequently to this change

  - All the "functional code" of the heuristics moved to `phys2bids.use_heuristics` rather than in the heuristic file, leaving it cleaner for the user (closes #44 ).
  - Instead of variables, the heuristic file populates a dictionary (this will be used to read yaml files too).
  - The documentation has been updated to reflect the changes.

Now, I have two doubts:
1. Is this a breaking change? It discontinues support for older heuristics, although not _that_ much.
2. I added a docstring to the heuristic file for completeness, but in this way the first code that the user should change comes after a long while - and a whole lot of information that probably is not necessary to them. I wouldn't see anything bad in removing the docstring in the heuristics to decrease the possibility of confounding the user, as in any case there's a full tutorial on how to use this function, we're not testing it, and we're not going to report it in the API.


phys2bids/phys2bids.py
```diff
@@ -137,39 +135,54 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
...
+        if `info['task']` is empty
     """
+    utils.check_file_exists(heur_file)
 
-    if sub[:4] != 'sub-':
-        name = f'sub-{sub}'
-    else:
-        name = sub
+    # Initialise a dictionary of info that has already "recording"
+    info = {'sub': '', 'ses': '', 'task': '', 'acq': '', 'ce': '',
```
eurunuela, line 147: `info` sounds a bit ambiguous to me. How about `bids_keys`?

phys2bids/heuristics/heur_euskalibur.py
```diff
@@ -0,0 +1,93 @@
+import fnmatch
```
eurunuela, line 1: I assume this one is an example?

smoia: Yep, why not.
it's the one used for a dataset that will be shared at a certain point.

smoia:
@eurunuela I addressed your point - hopefully also the coverage decrease.

@rmarkello @RayStick I had the same issue as #208 related to the deleted file. I think I solved it but @RayStick and @vinferrer please confirm that for our aims this old file: https://osf.io/u5dq8/ and this new file: https://osf.io/5829m/ are the same.

phys2bids/phys2bids.py
```diff
@@ -137,39 +135,54 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
...
+    if not bids_keys['task']:
+        raise KeyError(f'No "task" attribute found')
 
-    heur = utils.load_heuristic(heur_file)
-    name = heur.heur(Path(filename).stem, name)
+    # Compose name by looping in the bids_keys dictionary
+    # and adding nonempty keys
+    name = ''
+    for key in bids_keys:
+        if bids_keys[key]:
```
rmarkello, line 178:
```suggestion
        if bids_keys[key] != '':
```

I know that `''` evaluates to `False` but it's good practice to only use the `if X` notation if `X` is expected to be a boolean.

phys2bids/phys2bids.py
```diff
@@ -137,39 +135,54 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
...
-        name = f'sub-{sub}'
-    else:
-        name = sub
+    # Initialise a dictionary of bids_keys that has already "recording"
+    bids_keys = {'sub': '', 'ses': '', 'task': '', 'acq': '', 'ce': '',
+                 'dir': '', 'rec': '', 'run': '', 'recording': record_label}
 
-    fldr = os.path.join(outdir, name)
+    # Start filling bids_keys dictionary and path with subject and session
+    if sub[:4] == 'sub-':
```
rmarkello, line 151:
```suggestion
    if sub.startswith('sub-'):
```

docs/heuristic.rst
```diff
@@ -22,36 +22,34 @@ Initialisation
...
+| The list of possible keys is in the comment above, and corresponds to the list of possible entities of the `BIDs specification <https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/06-physiological-and-other-continuous-recordings.html>`_:
 
-- ``task`` stands for the name of the task. **It's the only required entity**, and it should match the task of the neuroimaging file associated to the physiological data.
+- ``task`` stands for the name of the task. **It's the only required entity**, and it should match the task of the neuroimaging file associated to the physiological data. If this is missing **the program will stop running**
```
rmarkello, line 48:
```suggestion
- ``task`` stands for the name of the task. **It's the only required entity**, and it should match the task of the neuroimaging file associated to the physiological data. If this is missing **the program will stop running**.
```

phys2bids/phys2bids.py
```diff
@@ -137,39 +135,54 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
...
+        bids_keys['sub'] = sub
+        fldr = os.path.join(outdir, f'sub-{sub}')
 
     if ses:
-        if ses[:4] != 'ses-':
-            ses = f'ses-{ses}'
-
-        fldr = os.path.join(fldr, ses)
-        name = f'{name}_{ses}'
+        if ses[:4] == 'ses-':
```
rmarkello, line 159:
```suggestion
        if ses.startswith('ses-'):
```

RayStick:
> @rmarkello @RayStick I had the same issue as #208 related to the deleted file. I think I solved it but @RayStick and @vinferrer please confirm that for our aims this old file: https://osf.io/u5dq8/ and this new file: https://osf.io/5829m/ are the same.

Apologies for that @smoia. I deleted them for a reason, but it was a stupid reason. Instead of explaining, I have actually just put those text files back now. They may be _slightly_ different lengths (i.e. different number of samples) compared to the original file, but otherwise they are the same. You can use either of those files, that you linked to (though 'Test3' will have have a new hyperlink now). Let me know if there are any issues.

smoia:
@RayStick, no worries! If it's shorter than the one we were using before and it has the same characteristics as the previous one that we use to test, then it's even better!
I just want to be sure.

@rmarkello, it's a great idea and we should implement it - not only for heuristics but also for mappings!
I would close this PR as is (with the suggestions you gave), and open an issue to add that feature (that will come extremely helpful once #206 is complete).
The only doubt that I still have is about the docstring in the heuristic file. Maybe it's not only not necessary, but also problematic. But if you think it's good as it is, think about which label to apply to this PR (if it's a breaking change or not - if it is the label is "majormod"), and merge it in!

smoia:
@rmarkello good ideas! I implemented them in the last commit.
I'm not totally sure about initialising the dictionary in the heuristic file vs passing it as an argument, but for sure updating the dictionary is a much better idea!

smoia:
Ok, codecov is not happy about the decrease in testing due to the new line that is in the function.
Since we don't have breaking tests (do we?), I would move on and merge it anyway, if @rmarkello you're happy with the latest version.

rmarkello:
This might now count as a breaking change... Let me know and I'll update the labels + merge in accordingly.

smoia:
@rmarkello I trust you - up to you to decide which label to add (majormod or minormod) and merge in!

PR #213, opened 2020-04-20
Add `getitem` and `eq` methods to BlueprintInput for easier slicing and comparing

smoia:
This PR is useful to complete #206.

## Proposed Changes

  - General update of outdated docstrings in physio_obj.py
  - Add attribute `trigger_idx` in `BlueprintInput` to indicate the trigger index
  - Add `__getitem__` method to slice `BlueprintInput` with multifreq
  - Add `__eq__` method for both classes, implemented in `are_equal()` function


phys2bids/physio_obj.py
```diff
@@ -162,17 +170,55 @@ def ch_amount(self):
...
+        sliced_timeseries = []
+
+        if isinstance(idx, int):
+            idx = slice(idx, idx + 1)
+
+        if not self.trigger_idx:
+            self.trigger_idx = 0
+
+        for n, channel in enumerate(self.timeseries):
+            idx_dict = {'start': idx.start, 'stop': idx.stop, 'step': idx.step}
```
sangfrois, line 199: So, for our purposes in multi-run processing : step will most likely be `0` right ?

phys2bids/physio_obj.py
```diff
@@ -162,17 +170,55 @@ def ch_amount(self):
...
+            for i in ['start', 'stop', 'step']:
+                if idx_dict[i]:
+                    idx_dict[i] = int(np.floor(self.freq[n]
+                                               / self.freq[self.trigger_idx]
+                                               * idx_dict[i]))
+
+            new_idx = slice(idx_dict['start'], idx_dict['stop'], idx_dict['step'])
+            sliced_timeseries[n] = channel[new_idx]
+
+        return BlueprintInput(sliced_timeseries, self.freq, self.ch_name,
```
sangfrois, line 209: That is excellent work by the way ; really opens up possibilities! I'll try and integrate these changes in split4phys function

phys2bids/physio_obj.py
smoia: Indeed! But this little function makes us be able to just have a standard indexing:
```
phys_in[230:360]
```

That's the main idea at least. Now I need to test it before we can actually use it!

smoia:
I believe this is ready for review.
I know that the `are_equal` function could probably be refactored A LOT, and I'd invite you to give me ideas on how to do so (or do it). If it won't take much to do it, then I can implement it in this PR, but please don't make it stop (too much) the PR merge because we need this for #206 . We can open a new PR for refactoring.

phys2bids/physio_obj.py
```diff
@@ -79,9 +79,93 @@ def has_size(var, data_size, token):
...
+                    return self == other
+                except ValueError:
+                    try:
+                        self['timeseries']
+                    except KeyError:
+                        return False
+                    except TypeError:
+                        return False
+                    else:
+                        return _deal_with_dict_value_error(self, other)
```
eurunuela, line 162: Please write some comments. It's not easy to follow this code without any comments.

phys2bids/physio_obj.py
```diff
@@ -162,17 +251,103 @@ def ch_amount(self):
...
+
+        Raises
+        ------
+        IndexError
+            If the idx, represented as a slice, is out of bounds.
+
+        Notes
+        -----
+        If idx is an integer, it returns an instantaneous moment for all channels.
+        If it's a slicing, it always return the full slice. This means that
```
eurunuela, line 280:
```suggestion
        If it's a slice, it always returns the full slice. This means that
```

phys2bids/physio_obj.py
```diff
@@ -162,17 +251,103 @@ def ch_amount(self):
...
+        potentially, depending on the frequencies, BlueprintInput[1] and
+        BlueprintInput[1:2] might return different results.
+        """
+        sliced_timeseries = [None] * self.ch_amount
+        return_instant = False
+        trigger_length = len(self.timeseries[self.trigger_idx])
+
+        if not self.trigger_idx:
+            self.trigger_idx = 0
```
eurunuela, line 289:
```suggestion
        if not self.trigger_idx:
            self.trigger_idx = 0
        
        trigger_length = len(self.timeseries[self.trigger_idx])
```

eurunuela: This should avoid getting errors in `self.timeseries[self.trigger_idx]`. See that if there's no `self.trigger_idx` that piece of code would raise an error.

phys2bids/physio_obj.py
```diff
@@ -162,17 +251,103 @@ def ch_amount(self):
...
+
+        # If idx is an integer, return an "instantaneous slice" and initialise slice
+        if isinstance(idx, int):
+            return_instant = True
+            if idx < 0:
+                idx = trigger_length + idx
```
eurunuela, line 295: Could you comment on this please?

phys2bids/physio_obj.py
```diff
@@ -162,17 +251,103 @@ def ch_amount(self):
...
+        other:
+            comparable object.
+
+        Returns
+        -------
+        boolean
         """
-        Renames the channels. If 'time' or 'trigger' were specified,
-        it makes sure that they're the first and second entry.
+        return are_equal(self, other)
```
eurunuela, line 339: Is this function really needed? It only calls the `are_equal( )` function. It does nothing else.

phys2bids/physio_obj.py
```diff
@@ -228,19 +401,22 @@ def delete_at_index(self, idx):
...
         """
         del self.timeseries[idx]
         del self.freq[idx]
         del self.ch_name[idx]
         del self.units[idx]
 
-    def check_trigger_amount(self, chtrig=1, thr=None, num_timepoints_expected=0, tr=0):
+        if self.trigger_idx == idx:
+            LGR.warning('Removing trigger channel - are you sure you are doing'
+                        'the right thing?')
```
eurunuela, line 414: Maybe we should wait for the user's input at this point and only proceed if the user confirms.

phys2bids/physio_obj.py
eurunuela: Okay, I see now this is part of the class and the `are_equal( )` function isn't. I'm sorry for that. You know how GitHub hides stuff in reviews.

phys2bids/physio_obj.py
smoia: That could be an option, but then if you run a batch of files they won't run smoothly - if there's a reason for the user to remove the trigger channel.
Why don't we add the possibility of such interaction in a different PR? We can add a parser argument that would pause the workflow in particular cases.

smoia:
@eurunuela I added comments, and in doing so I realised part of the code could have been collected in a local definition. I also tested it locally and all the checks passed (although I had to rerun it due to the same issue reported in #221 ).
If you want Travis CI to work, we can think about having a different host for the files, at least momentarily. Let me know.

phys2bids/physio_obj.py
eurunuela: That's what I was thinking actually; i.e. to add an argument to the parser. However, I'd do it so that it stops by default. It would be like giving the "pro" option to power-users.

What do you think?

smoia: Sure, agreed on having an option! Can you open an enhancement issue about it?
Don't mark it for the beta milestone though - we have enough stuff in it.

eurunuela:
Just ran the tests. There was an error on `test_txt.py`:

`urllib.error.ContentTooShortError: <urlopen error retrieval incomplete: got only 8944344 out of 8953000 bytes>`

Given that this PR is not labeled as urgent, I'd wait until we close #221 and #215. 

We can discuss how to proceed during our meeting on Thursday if that's okay with you. I'd like to get @rmarkello 's opinion on this too.

smoia:
We do need this PR for #206, but we can wait until tomorrow and discuss what to do with the tests first!

smoia:
I'm sorry to bother you @eurunuela @vinferrer @sangfrois , but we need this PR to safely move on with #206 . Could you please review it, now that the tests are back working and are passing?
Thank you!

PR #219, opened 2020-05-08
Add multi-run workflow synchronized with BIDS dataset

### LINKED ISSUE #217: Method to split multi-run physio files using BIDS dataset or dicom directory
In order to minimize user input and to work around missing physio triggers, it would be nice to have a function that splits and labels physio files using scan information from either the converted BIDS dataset or the raw dicoms.

## Detailed Description

The method I'm proposing would compare complete information from the actual scan data to potentially incomplete information from a multi-run physio file. It would use the difference in time between each scan's onset and each physio trigger period's onset to determine which scan matches up with each trigger period. From there, the timing of the scans can be used to split the physio file, even in the case of missing trigger periods. Moreover, because the _relative_ timing within each file is what matters, this could be used even when (1) scan timing is anonymized, (2) scanner and physio computer clocks are mismatched, or (3) the physio file lacks absolute timing.

## Context / Motivation

In my experience, BioPac can sometimes randomly fail to register trigger signals from PsychoPy or E-Prime (or possibly the behavioral tasks simply fail to send the trigger signals properly). Either way, my own physio files can have either missing scan-associated trigger signals or signals that go on longer than the scan (e.g., when the task crashes or is stopped early). In these cases, the proposed method could be used instead of manually editing the physio files. Plus, it would require minimal information from the user.

## Possible Implementation

The scans.tsv file in BIDS datasets tends to have an `acq_time` field with the date and time of the start of the scan down to the second. The json metadata file associated with each scan also tends to have an `AcquisitionTime` field with the time of the start of the scan down to the millisecond. Dicoms of course also generally have the same information, so they could be scraped in cases where the dataset isn't available. From these files, we can get the onsets of all of the scans w.r.t. one another.

We can also take a multi-run physio file and extract the relative onsets of each of the _recorded_ scans. As long as there are a couple of successful trigger signals, and the onsets of those signals are accurate, then we can compare the timing of the physio trigger signals to the timing of the scans. Assuming that there is variability in the timing between scans (e.g., every scan doesn't start exactly X minutes after the last one started), we can determine the offset that should be applied to the scan times in order to get accurate physio periods. 

Plus, if we use the scans.tsv file, we can just grab the actual filenames for the different scans when we split the physio!

See [here](https://github.com/tsalo/misc-notebooks/blob/13557e3145a9da647a9d8e2e09819593a866dbd2/compare_physio.ipynb) for a very basic proof of concept.

NOTE: The only part I'm still struggling with is inferring the durations of the scans for splitting the physio files. If we restrict this function to functional data, we can just load up each scan's TR and number of volumes, and then multiply those to get the duration. If we use the dicom directory, it could be a bit more difficult.

tsalo:
Closes #217. This enhancement adds several functions to do the following:
1. Load relevant (currently just BOLD) scans from BIDS dataset and determine acquisition time, duration, and generalized name for each.
2. Extract trigger onsets from `BlueprintInput` object.
3. Attempt to match physio trigger onsets to scan onsets.
4. Interpolate any missing physio trigger periods from available scans.
5. Identify onsets and offsets for physio trigger periods for each frequency in the physio object.
6. Slice the physio object into acquisition-specific `BlueprintOutput`s.
7. Save `BlueprintOutput` objects to BIDS files.
8. Generate a figure showing timing of scans and physio trigger periods before and after synchronization, for QC purposes.

## Current Development/Testing Strategy
1. Requires a BIDS dataset and an associated physio file.
2. Do the following:
```python
from phys2bids.conversion import workflow
from phys2bids.interfaces.acq import populate_physio_obj

bids_dir = '/path/to/bids/'
subject = '01'
session = '01'
physio_file = '/path/to/physio/file.acq'
physio = populate_physio_obj(physio_file, chtrig=3)
res = workflow(physio, bids_dir, sub=subject, ses=session)
```
3. Review output filenames and `BlueprintOutput` objects.

## To Do
- [x] In dicoms, `SeriesTime`, `InstanceCreationTime`, and `ContentTime` are very similar, but `AcquisitionTime` is generally ~20 seconds earlier than the rest. ~~We need to determine which time is closest to the scanner's trigger pulse.~~ AcquisitionTime is when the trigger pulse is sent, but heudiconv uses ContentTime.
    - Change proposed in https://github.com/nipy/heudiconv/issues/450.
    - For now, I'm comfortable just loading the json to get `AcquisitionTime` instead of waiting for `acq_time` in the scans file. Checking this one off.
- [x] Heudiconv-generated BIDS datasets retain the sub-second resolution for acquisition time in the sidecar json files, but the scans.tsv file only has second-level resolution. We need sub-second resolution, so we should raise an issue about retaining that resolution in the scans.tsv files to make it easier to extract accurate onsets.
    - Change proposed in https://github.com/nipy/heudiconv/issues/447 (with several additional issues and PRs spread throughout heudiconv, bids-specification, and bids-validator).
    - UPDATE: Associated PR has been merged.
- [x] There is some variability in the delay between any of the possible scan onset times and the physio trigger periods. I assume this is lag from whatever task is running, and the physio triggers are the ones that are mistimed.
    - Handled in 1b198c453bf51abb50ada6aaf9d0d6aef3a2c174.
- [X] Missing scans or physio triggers make it harder to find the correct time offset. We will need to loop through both scans (in case first scan's physio trigger is missing) and physio triggers (in case first physio trigger's scan was dropped during conversion), and somehow select the best match to determine the correct delay.
    - Also handled in 1b198c453bf51abb50ada6aaf9d0d6aef3a2c174.
- [x] Use `phys2bids` classes and methods for combining and splitting physio data.
    - I have given up on combining discrete segments from physio data for this PR, and will assume that each physio object represents one continuous recording.
- [x] Merge onsets of multi-contrast runs in scans file, either using the acquisition times (which are very close) or the entities in the filenames.
- [x] Infer output filenames from scans file, accounting for multi-contrast entities like `echo`.
- [ ] Conceptual review
- [ ] Integrate figure into reports
- [ ] Tests

## Proposed Changes
  - Add a completely (hopefully) unsupervised method for splitting multi-run physio files using onset-time and duration information from a converted BIDS dataset.


pvelasco:
> 1. In dicoms, `SeriesTime`, `InstanceCreationTime`, and `ContentTime` are very similar, but `AcquisitionTime` is generally ~20 seconds earlier than the rest. We need to determine which time is closest to the scanner's trigger pulse.

From some documentation I found from Siemens on image reconstruction:

> AcquisitionTime: The Time the acquisition of data started.

From an answer given by Chris Roden ([dcm2niix](https://www.nitrc.org/forum/message.php?msg_id=29409)):
> Instance Creation Time 0008,0013 initially reflects the time an image leaves a parallel processing reconstructor, so it has a variable delay with regards to the acquisition time. Acquisition Time is more meaningful for data.

So, if the reconstruction involves something more than a straight FFT (i.e., GRAPPA, SENSE, etc.), that will explain the ~20 sec delay you find between the `AcquisitionTime` and the other times.

If I remember correctly, for Siemens scanners, the "scanner trigger" is sent at the time of the excitation pulse of the first slice of each (collected) volume.  So the very first trigger will be sent just a few milliseconds before the first chunk of data for the first slice for the first volume is acquired (`AcquisitionTime`).

Regarding the time resolution (seconds vs. sub-second): I think the purpose of `func_times.tsv` and `physio_times.tsv` is to know which physio file (or chunk of a session-long one) corresponds to which functional file, isn't it?.  For that, second-level resolution should be enough.  Then, the fine alignment (the one requiring sub-second resolution) should be handled by looking at the first trigger signal in each physio file. That should be your "Time zero", according to BIDs.

tsalo:
Thanks @pvelasco! I'm glad to know that AcquisitionTime is the right one, although it looks like heudiconv [grabs acq_time from ContentTime instead of AcquisitionTime](https://github.com/nipy/heudiconv/blob/75f2850f7f07a95ff5875aba1dfd76c09e4be201/heudiconv/bids.py#L408-L411). It might be best to use AcquisitionTime instead, although I can follow up about that in https://github.com/nipy/heudiconv/issues/447.

> Regarding the time resolution (seconds vs. sub-second): I think the purpose of func_times.tsv and physio_times.tsv is to know which physio file (or chunk of a session-long one) corresponds to which functional file, isn't it?. For that, second-level resolution should be enough. Then, the fine alignment (the one requiring sub-second resolution) should be handled by looking at the first trigger signal in each physio file. That should be your "Time zero", according to BIDs.

Yes, the scan times are used to find the trigger periods that correspond to each functional file, but this approach won't work for single-scan physio files, because it relies on the exact times _between_ physio trigger periods as well as those between scan onsets.

At the moment, my code finds the trigger signal that would have the shortest delay from the scan acquisition time, rather than the first trigger signal. My logic is that, since the triggers occur at variable delays, the one with the shortest delay is the one that is closest to the _actual_ start of acquisition, and is thus the one we should build our timing information from. To find that time, we need scan times that are as accurate as possible. Well, actually we need relative onsets to be accurate. The absolute scan times are not used, because I didn't want to assume that the physio and scan computers' clocks were perfectly synchronized, and because the BioPac segment timestamps seem to only have second-level resolution.

tsalo:
The functions are now in a state that I can call them sequentially in a notebook (see [here](https://github.com/tsalo/misc-notebooks/blob/2def00b64c7377155eb78bb8945f23ffecb3fecc/sync_physio.ipynb)). I have also made some time series figures of the physio and functional timing from the results (see [here](https://github.com/tsalo/misc-notebooks/blob/2def00b64c7377155eb78bb8945f23ffecb3fecc/plot_timing.ipynb) for the notebook and below for the figures).

You might notice a few odd things below:
- Some of the blue (physio) blocks extend past the red (scan) blocks. This is because my task failed to reset the trigger when it ended, so the trigger is only reset _right_ before the next task starts, by that next task.
- Some blue blocks that don't have corresponding red blocks. That's where scans were dropped. In this case, the first task was stopped and restarted, as was the second one.
- It's hard to see here but the last block has a slight mismatch between the red and blue blocks. Our last task actually sends the trigger about 6 seconds late, which is something I was not aware of until I started working on this code!

timeseries_unsynchronized
timeseries_synchronized

phys2bids/conversion.py
```diff
@@ -0,0 +1,310 @@
...
+    Collect onsets from physio file, both in terms of seconds and time series
+    indices.
+    TODO: Replace file-loading with phys2bids physio_obj
+    TODO: Stitch segments together before extracting onsets.
+    """
+    import bioread
+    from operator import itemgetter
+    from itertools import groupby
+    physio_data = bioread.read_file(physio_file)
+    trigger_channel = physio_data.channels[-1]
```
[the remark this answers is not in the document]
tsalo: One of the things on my to-do list is to use the native `phys2bids` classes. I don't know much about them at the moment, so that may take some time, but I believe that selecting the trigger channel is one of the possible parameters for the main physio container object.

That said, if your interest is more in making it easier to test/debug for the near future, I'm happy to make the trigger channel selection more flexible.

tsalo:
There are no tests yet, but I think this is ready for review.

tsalo:
@smoia- I have a question about desired functionality. This method should be able to correct inaccurate trigger signals. Would you prefer to retain the original trigger signal in each sliced object (to keep fidelity with the original data) or to use a corrected version?

Also, should the trigger channel be retained in each frequency-specific `BlueprintOutput` object, perhaps with resampling?

eurunuela:
I'd lean towards keeping the original trigger signal so it makes our lives easier when debugging.

tsalo:
@eurunuela Thanks! Retaining the original trigger time series will reduce or remove any confidence we have in the triggers for split files, but I'm honestly not sure that that matters, as long as the time offset is correct (which it should be with the workflow). Plus, I agree that keeping the converted data as close as possible to the raw data is for the best, as long as an accurate trigger time series _isn't_ super important.

tsalo:
@eurunuela by the way, this workflow will generate a figure like the one from earlier in this PR's thread. It would be great to incorporate it into the reports at some point.

smoia:
@tsalo we can discuss about it with the others during the upcoming meeting, but I prefer to maintain the original data as much as possible too. However, if there is a good reason to think that a) we need the trigger signal for other applications and the information it can offer does not overlap with other properties (e.g. the acquisition starting time), another possibility we can implement is to "backup" the original trigger in another channel and substitute the original one with a patched version - or even better add a patched version of the original trigger in a new channel.

eurunuela:
@tsalo thanks! Sure! We can work on adding the plot to the reports. I personally would merge #243 first though.

eurunuela:
@smoia I was thinking we could backup the original trigger with another attribute of the BlueprintInput class or whatever we call that now.

smoia:
@eurunuela  what would be the purpose of backing it up in another attribute?

IMHO, if we want to have a patched version of the trigger, we also want the original trigger to be in the output files. So I would say it's sufficient to have it between the channels. I might be missing something though.

smoia:
Oh, and btw, yes @tsalo, we should have the trigger in every output file. I'm going to open an issue for that!

eurunuela:
> @eurunuela what would be the purpose of backing it up in another attribute?

To have a "clean" output but at the same time have the option to debug/compare. I don't know if it makes sense. You work with physio data much more than I do so maybe it makes sense for you to have the original one in another channel.

smoia:
> > @eurunuela what would be the purpose of backing it up in another attribute?
> 
> To have a "clean" output but at the same time have the option to debug/compare. I don't know if it makes sense. You work with physio data much more than I do so maybe it makes sense for you to have the original one in another channel.

Right. If the final solution is not to save the original signal, then I agree that should be the way to go, yes.
@tsalo, both @eurunuela and I lean toward saving the original file - but I like your idea of adding a clean trigger in case the original one has something wrong, maybe in a new channel? What's your idea?

tsalo:
From what I can tell, the trigger signal may be inaccurate in the following ways:
1. Delayed start: The trigger onset will vary slightly due to task loading stuff. It's just the nature of tasks. Additionally, bugs may cause longer trigger delays (as when I noticed that one of my project's tasks has a ~6 second delay!).
2. Delayed end: If the task crashes or simply fails to send the trigger reset signal, then the trigger won't end until the start of the _next_ scan. It shouldn't be a problem for the physio data, though, since we tend to keep data from _before_ the trigger, rather than after.
3. No signal: If a task fails to send the trigger signal (as seems to happen randomly for at least one of my project's tasks), then there will be no trigger signal at all for that run.

The original trigger could go in a new channel called "raw_trigger" or something, I guess. Since the channel label "trigger" is written into the specification, I think that, if we need a corrected trigger signal at any point, then it should be the one with the label "trigger". If we don't need the trigger signal for anything, then it doesn't really matter IMHO.

smoia:
@tsalo, I see your point, but that's a particular case, not the norm - for instance, if you sample the trigger directly from the MRI then I would argue it's independent from the task script.
The way you'd implement this function would you do it every time, independently from the file, or only if the algorithm finds something is wrong/the user asks for this option?

I think it's worth discussing in the meeting. Can you add it to the meeting agenda pls?

tsalo:
> for instance, if you sample the trigger directly from the MRI then I would argue it's independent from the task script.

Yes that's very true. But then, in that case, the trigger signal would never be inaccurate, so the "corrected" signal and the original signal would be the same. There's no reason to back up the signal in that case.

> The way you'd implement this function would you do it every time, independently from the file, or only if the algorithm finds something is wrong/the user asks for this option?

The correction is only possible when you have the timing information from the BIDS dataset, in which case we have a "ground truth" (the imaging data timing) and an "estimate" (the recorded trigger signal). In that case, I would automatically estimate the corrected trigger signal (i.e., no user input needed). Users will need to do visual QC using the figure though.

> I think it's worth discussing in the meeting. Can you add it to the meeting agenda pls?

Okay, will do.

smoia:
> > The way you'd implement this function would you do it every time, independently from the file, or only if the algorithm finds something is wrong/the user asks for this option?
> 
> The correction is only possible when you have the timing information from the BIDS dataset, in which case we have a "ground truth" (the imaging data timing) and an "estimate" (the recorded trigger signal). In that case, I would automatically estimate the corrected trigger signal (i.e., no user input needed). Users will need to do visual QC using the figure though.

Ok, but then I would do this operation only if the two versions of the trigger are not the same - or if the user asks for it.

phys2bids/conversion.py
```diff
@@ -0,0 +1,403 @@
...
+    df = determine_scan_durations(layout, df, sub=sub, ses=ses)
+    df = df.dropna(subset=['duration'])  # limit to relevant scans
+
+    # Convert scan times to relative onsets (first scan is at 0 seconds)
+    df['onset'] = (df['acq_time'] - df['acq_time'].min())
+    df['onset'] = df['onset'].dt.total_seconds()
+    return df
+
+
+def determine_scan_durations(layout, scan_df, sub, ses=None):
```
tsalo, line 81: @RayStick I would love to get your input on this function. It makes the very simple assumption that scan duration = number of volumes * repetition time, but I don't know if that holds for ASL. We could also load and look at [`AcquisitionDuration` and other fields](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#timing-parameters_1), if that would help.

phys2bids/conversion.py
```diff
@@ -0,0 +1,403 @@
...
+
+    Returns
+    -------
+    df : pandas.DataFrame
+        DataFrame with one row for each trigger period and three columns:
+        onset (in seconds), index (onsets in index of time series array),
+        and duration (in seconds).
+    """
+    samplerate = 1. / freq
+    scan_idx = np.where(trigger_timeseries > 0)[0]
```
tsalo, line 143: This requires the trigger to be binarized already.

phys2bids/slice4phys.py
```diff
@@ -107,6 +108,71 @@ def find_runs(phys_in, ntp_list, tr_list, thr=None, padding=9):
     return run_timestamps
 
 
+def slice_phys(phys, run_timestamps, time_before=5):
```
tsalo, line 111: @sangfrois Would you mind taking a look at this? My goal is to separate the slicing step from calling `find_runs`, since my workflow doesn't use that function. The modularization will also aid testing.

phys2bids/conversion.py
```diff
@@ -0,0 +1,403 @@
...
+       scans that weren't kept in the BIDS dataset.
+    7. Split physio file based on scan-specific onsets and offsets.
+    8. Write out scan-associated physio files.
+    9. Generate and return scan/physio onsets figure for manual QC.
+    """
+    layout = BIDSLayout(bids_dir)
+    scan_df = load_scan_data(layout, sub=sub, ses=ses)
+
+    trigger_timeseries = physio.timeseries[physio.trigger_idx + 1]
+    freq = physio.freq[physio.trigger_idx + 1]
```
tsalo, line 381: TODO: Drop `+ 1` once #275 is merged.

phys2bids/conversion.py
```diff
@@ -0,0 +1,403 @@
...
+        split_times = (row['index_onset'], row['index_offset'])
+        run_dict[base_fname] = split_times
+    phys_dict = slice_phys(physio, run_dict, time_before=6)
+    outputs = []
+    for k, v in phys_dict.items():
+        output = BlueprintOutput.init_from_blueprint(v)
+        output.filename = k
+        outputs.append(output)
+        # Let's not actually save files until we're more confident.
+        # output.save()
```
tsalo, line 402: TODO: Uncomment before merging.

smoia:
Continuing our chat about trigger correction vs non-correction, I need to see this code in action to have a definitive opinion about it, but I still lean toward saving and outputting the original trigger signal as default behaviour. Once again, if we record the trigger directly from the MRI, the channel might be more accurate than any other information we can use to guess the beginning or end of a scan. I still need to go through the code, but if you use the time of creation of a file (or a recorded clock time), then if two different machines (e.g. one recording the MRI data and the other recording BIOPAC signals) have different times, we risk to mis-align the trigger while applying the correction. Similarly, if you are binarising your trigger as scan/not-scan, you're taking away a  lot of information that could be useful downstream.
I would invite you to add a flag in the parser to let the user ask for this correction (expecially if you are overwriting the channel) and/or to store this correction in a different trigger channel.

@RayStick any opinion on this?

tsalo:
> I would invite you to add a flag in the parser to let the user ask for this correction (expecially if you are overwriting the channel) and/or to store this correction in a different trigger channel.

@smoia I can do that.

phys2bids/conversion.py
smoia: Ok, I have to say I didn't have a look at the code yet, but what do you mean by this?
In which format do you need the trigger representation? Do you need a "block" timeseries that is `True`/`1` when the sequence is going on and `False`/`0` when it's not?

tsalo: Currently, it does require those blocks, but I think we can make it flexible enough to work with other formats.

Truthfully, what I really need is onsets and durations of trigger block periods in terms of seconds and array indices, so we could move the step that actually gets those to outside of the workflow if that would be easier.

smoia: Ok, gotcha. The onset of the trigger block is a property of the classes we use throughout the pipeline (I don't remember its name now but it's there!). At the moment it's an integer because we didn't need one for each multirun block, but we can adapt it to be a dictionary and contain all of them - although that also needs to adapt the code in `split4phys`. The duration can be obtained by the `ntp*tr` variables, that are the number of timepoints expected and their length. We can add those as properties of the class as well if it helps.

Alternatively, we can add those "timeblocks" as either the last channel in the timeseries, or as another property of the class to keep the timeseries "clean". We would probably need to adapt the code in `split4phys` anyway to do create that timeblock, but it's also a good way to go IMHO.

phys2bids/conversion.py
RayStick: _Most_ of the time this is the case for ASL outputs, but sometimes it isn't. I think, for now, it's fair to assume this (duration=vols*reps), and then we work on making it more adaptable later? In the funny cases where there is one long TR at the start (i.e. M0scan or 'calibration' scan proceeding the main ASL scan) it might be simplest if the user splits these two files first but I am not fully sure - I'd need to think about it a bit more.

RayStick:
Hi all, I have read through what people have written but I have done so rather quickly (so as to catch up), so sorry if I misunderstand things.  

  In general, I agree with @smoia - I think I will need to see this code in action to understand it properly. I also lean towards have an “old triggers” column and a new “corrected” triggers column, as it probably helps for troubleshooting.   In general, this PR looks like its adding some great functionality, bu it may only be relevant for certain experimental and equipment set-ups? I will explain a bit about my experimental set-up, below, and you can let me know if you agree! 

**Rationale/Concept** - I don’t think I have had a time where I have had just a few missing triggers in my physiological recordings, however I have had occasions where I have not recorded any trigger information at all, alongside the physiological data (due to forgetting to plug the trigger connection in ... ) which is very annoying. From what I have picked up, I think I have quite a different set-up to you @tsalo. I don’t have any synchronization between task (stimuli presentation) and trigger recording alongside the physiological data. I just record all the volume triggers that the scanner sends during fMRI sequences, but this does not trigger anything. (Sometimes I get the the stimuli to be triggered by the first volume trigger of a scan, but that is separate from the physiological recording equipment).  

**Implementation** -  I am struggling to work out how this would exactly apply for the type of data I collect, so I will explain what we do. We first turn on our physiological recording software a while before we start any scanning, as it takes time to attach all the peripheral devices to a person, and check the signals look reasonable. We then record continuously throughout the whole MRI session i.e. all physiological data from one MRI session is contained within one file. Within this  recording of physiological data, we record all MRI volume triggers. Therefore the start time of the physiological recording file is not synched at all with any other start time (task, MRI scan). However, I understand that this code is going to look at the _relative_ timings between MRI scans to work out the relative physiological timing blocks? That makes some sense to me, but does it therefore always assume you have the first trigger of an fMRI scan, to get the onset of this block correct? And then if there are any triggers missing in the middle it can fill them in just fine? Otherwise, I am struggling to work out how this works without having a trigger recorded at the onset of an fMRI scan.

tsalo:
> In general, this PR looks like its adding some great functionality, but it may only be relevant for certain experimental and equipment set-ups?

The ability of the workflow to correct trigger time series is only relevant for some setups, and the workflow will most definitely not work with single-run setups, but the general ability of the workflow to identify and label physio runs associated with specific imaging runs is broadly useful. Basically, you shouldn't have to feed in any information beyond a path to the BIDS dataset and the subject/session to get your physio data.

> I am struggling to work out how this would exactly apply for the type of data I collect

Your trigger time series is going to be accurate (unlike mine), but it's also completely unlabeled. You still need to use a heuristic or something to identify which block of physio data corresponds to which imaging file in the BIDS dataset. This should automate that.

> Therefore the start time of the physiological recording file is not synched at all with any other start time (task, MRI scan). However, I understand that this code is going to look at the relative timings between MRI scans to work out the relative physiological timing blocks? That makes some sense to me, but does it therefore always assume you have the first trigger of an fMRI scan, to get the onset of this block correct? And then if there are any triggers missing in the middle it can fill them in just fine? Otherwise, I am struggling to work out how this works without having a trigger recorded at the onset of an fMRI scan.

You need starts of physio trigger periods to generally match up with starts of fMRI scans for this to work, but a lot of the extra stuff I put into the workflow is to make sure it will work in cases where the first scan trigger or imaging data are missing.

Think of it like this. You have two arrays of "onsets" with the same scale (seconds) but different start points. You know the durations and "names" for the different onsets for one of the arrays, but not the other. Here's some pseudocode to illustrate:

```python
physio_onsets = [1.5, 7, 12.2, 22, 27]  # from physio file
scan_onsets = [0.8, 6.3, 10.7, 20.5, 25.5]  # from BIDS dataset
scan_durations = [1, 2, 3, 4, 5]  # from BIDS dataset
scan_names = ['a', 'b', 'c', 'd', 'e']  # from BIDS dataset
# the times between the onsets at the same
delta_physio = np.diff(physio_onsets)  # [5.5., 5.2, 9.8, 5]
delta_scans = np.diff(scan_onsets)  # [5.5., 5.2, 9.8, 5]
delta_physio == delta_scans  # True
# So now we find what the scan onsets (and durations) would be
# in terms of the physio signal.

# I'll estimate the onset from just the first scan and physio block
# because we *know* neither is missing in this case
offset = scan_onsets[0] - physio_onsets[0]  # -0.7
new_scan_onsets = scan_onsets + offset
# now we loop through, take each new scan (onset : onset + duration)
# and name the file based on the scan name.
```

There were three things I noticed in my own data that complicated matters:
1. Physio trigger onsets happens after a delay, at variable amounts. The delay is a natural consequence of using the task to send the signal, and is generally quite small (except for a bug in one of my tasks). The variability is also quite small and also a consequence of using an intermediary.
    - This does not impact your data, based on your setup.
    - It's nice to be able to "correct" the trigger signal in these cases, since the task-relayed trigger signal is inaccurate.
2. Some physio trigger periods were missing, and thus the scans had no corresponding blocks to compare against. This happened sometimes when my tasks failed to send the signal to BioPac.
    - This does not impact your data, based on your setup.
    - It's nice to be able to "correct" the trigger signal in these cases as well, since the task-relayed trigger signal is, well, completely missing.
3. Some scans were "missing". Sometimes, you stop a scan early or don't convert it over to the BIDS dataset because it's just that terrible. Whatever the reason, the BIDS dataset won't always reflect _all_ of the physio trigger periods.
    - This impacts everyone's data, and is the main reason why we don't want to assume that the first physio trigger matches the first scan in the dataset.

To deal with these three things, my method is somewhat more complicated than the pseudocode above. Basically, I take the onset delta arrays for the physio and scan data and find the "offset" that minimizes the overall difference between the two arrays. In my data, that overall difference will be > 0. In your data (limited to cases where _all_ scans were retained in the BIDS dataset), that difference should be 0.

Of course, the fewer scans there are and the more missing runs there are (both in physio and BIDS), the more likely the method is to be wrong, so that's why it outputs a figure as well.

phys2bids/conversion.py
tsalo: > Alternatively, we can add those "timeblocks" as either the last channel in the timeseries, or as another property of the class to keep the timeseries "clean".

I like the idea of a binarized version of the trigger time series, perhaps in a separate property. That way I could just assume that this property contains a boolean array and work from there. We just need to ensure that the binarization procedure works for both block-like and variable trigger signals.

RayStick:
Thanks for your detailed reply. 
Before I reply properly, I have a quick question - how do you define the physio_onsets from the physio file?

tsalo:
I extract contiguous blocks of non-zero values in the trigger signal and identify the time of the first point in each block. 

This step can be divorced from the main workflow code, though, in order to support your type of trigger signal. Per https://github.com/physiopy/phys2bids/pull/219#discussion_r457513658, one good route might be to binarize and "blockify" the trigger signal before the workflow, and retain that simplified trigger in a separate channel. Then the workflow could use the block-form trigger without touching the original trigger.

RayStick:
> I extract contiguous blocks of non-zero values in the trigger signal and identify the time of the first point in each block.
> 
> This step can be divorced from the main workflow code, though, in order to support your type of trigger signal. Per #219 (comment), one good route might be to binarize and "blockify" the trigger signal before the workflow, and retain that simplified trigger in a separate channel. Then the workflow could use the block-form trigger without touching the original trigger.

From re-reading your last comment, and some of the in-line comments, I think I follow it all now. Thanks!

phys2bids/conversion.py
tsalo: I've opened #278 about this. It will need to be address before this can be merged, and I think it merits a separate PR given that handling variable triggers is outside my area of expertise.

RayStick:
Shall I wait for conflicts to be resolved before properly reviewing?

tsalo:
@RayStick I've resolved the conflicts so it's good to go. Thanks!

smoia:
@tsalo I'm sorry but I really can't see if there was something that I missed and I should have done for this PR. Is it "only" missing new reviews? If so, Maybe someone else could take the lead on this.

What is the final solution on the trigger channel? Are you rewriting it or adding another "corrected" or "binarised" trigger channel?

tsalo:
@smoia See this thread: https://github.com/physiopy/phys2bids/pull/219/files#r455898174. Basically, the method will only work on trigger signals that are already block-ish. I didn't implement a solution for that in this PR, but I did open #278.

RayStick:
I will take a look on Monday, though bear in mind I have not looked at the phys2bids code for a while now ...  so I might be a bit slow to review this! If it needs to be merged sooner, I'd suggest someone else jump in as second reviewer, sorry!

smoia:
Hey @RayStick ! Thank you, but I would wait a tad longer. This PR is not passing any test, and it depends on #278 .
In fact, I'd revert it to draft @tsalo .
There are other reviewes needed if you're eager to do one ;)

Issue #185, opened 2020-03-24
Check that phys2bids is BIDS 1.3.0 compliant.


smoia:
## Detailed Description

At the moment, the latest stable BIDS is `v.1.3.0`.
BIDS has detailed specifications on many things, starting with how the files should be named, how they should be organised internally (see #184), how  information should be reported (see #127).
I'm not totally sure that `phys2bids` complies to that specification - although we definitely have to!

## Possible Implementation

1. Read BIDS 1.3.0, especially the section relative to physiological files, functional MRI files, units of measure.
2. Check that our output is compliant to that specification
3. If not, open bug issues to have it fixed
4. Add in the documentation and in the `readme` that we are bids 1.3.0 compliant (a sentence? a nice badge?)

smoia:
@RayStick @rmarkello , do you think this could be a good first issue?

smoia:
@vinferrer at what point is this issue? Do you need any help?

vinferrer:
I didn't had time to go through it. Help is always welcome, specially since i am no expert in BIDS and I am bad at details. Which is important for standards

vinferrer:
In theory #222 should close the units of measure part. Although ``mmHg`` is a cranky issue since is not bids standard but is necessary in some channels.

vinferrer:
I don't understand in which way specifically functional MRI files standard affects us. I suppose it is in an indirect manner

vinferrer:
I tried the tutorial file with the heuristic and the names and structure seem to check with bids standard

vinferrer:
However in this case there is not stimuli file, don't think we have those

vinferrer:
Lastly, regarding the JSON file, all seems good to me. However, I was wondering what happens with the sampling frequency when you have multifrequency files. Does that generate different ``tsv and JSON`` files for each frequency?

smoia:
> Lastly, regarding the JSON file, all seems good to me. However, I was wondering what happens with the sampling frequency when you have multifrequency files. Does that generate different `tsv and JSON` files for each frequency?

Indeed, as it's the expected output to have  BIDS compliant physiological files (from what I understood from the BIDS specs).

> However in this case there is not stimuli file, don't think we have those

We don't for the moment. I don't know if we want to support Respiract files, in which case we might have stimuli files? @RayStick @BrightMG what do you think about this?

> I don't understand in which way specifically functional MRI files standard affects us. I suppose it is in an indirect manner

[...]

PR #197, opened 2020-04-02
Add guidelines for PR reviews.

smoia:
Closes nothing, but follows #181
I tried to write down some guidelines for PR reviewers (and update those for PR authors), as that process could follow guidelines too.
Before even reviewing it, I'd like @rmarkello @eurunuela @vinferrer @RayStick and the rest of the team to throw in more suggestions and discuss these guidelines.

## Proposed Changes

  - Add review guidelines in contributor guide.
  - Minor adjustments related to the adoption of `auto`

PR #222, opened 2020-05-22
Add BIDS-compatible automatic reformatting of units of measure

### LINKED ISSUE #127: Adhering to BIDS units standards
## Detailed Description

[BIDS uses a standard format for units of measure too](https://bids-specification.readthedocs.io/en/stable/99-appendices/05-units.html).
Our files have different ways to report units, but we should treat them with phys2bids!

## Possible Implementation

A possible implementation is adding a dictionary and a function, either in the main script or in a separate file, that gets called toward the end of the `_main()` function - before exporting the `json` file.
The dictionary contains different possible types of homonyms for the units, and the function reads what units our objects contain and changes them accordingly to the dictionary.

vinferrer:
Closes #127 

## Proposed Changes

  - add bids_untits.py 
  - This file contains dictionaries and fucntions to convert units strings to proper bids formatting.
  - bidify_units converts unit string to bids formatting
  - this function doesn't do unit conversion it only puts the unit string in the right format


phys2bids/bids_units.py
```diff
@@ -0,0 +1,94 @@
...
+def bidsify_units(orig_unit):
+    """
+    Read the input unit of measure and use the dictionary of aliases
+    to bidsify its value.
+    It is possible to make simple conversions
+
+    Parameters
+    ----------
+    unit: string
+        Unit of measure, might o might not be BIDS compliant.
```
eurunuela, line 54:
```suggestion
        Unit of measure, might or might not be BIDS compliant.
```

phys2bids/bids_units.py
```diff
@@ -0,0 +1,94 @@
...
+    the other for prefixes (e.g. "milli"). However, that is going to be tricky,
+    unless there is a weird way to multiply two dictionaries together.
+    """
+    # call prefix and unit dicts
+    new_unit = ""
+    # for every unit alias in the dict
+    for u_key in unit_aliases.keys():
+        unit = orig_unit.lower()
+        # check that u_key is part of unit
+        if u_key in unit:
```
eurunuela, line 73: Wouldn't it be easier to just do `if orig_unit.lower() in unit_aliases.keys():` or someting like that? (I'm quite sure this doesn't work but you get the idea)

My point is, there's got to be a way to avoid iterating over the entire list of aliases.

smoia:
@eurunuela you automatically volunteered to be the main reviewer of this PR!
I can't be second reviewer though. @rmarkello could you?

PS: I'd like to see the tests not failing for another couple of times, but yes, it looks like all the issues were solved! Still, it would be great to have lighter files and faster tests, so let's keep doing that!

eurunuela:
> @eurunuela you automatically volunteered to be the main reviewer of this PR!
> I can't be second reviewer though. @rmarkello could you?

Well, I was requested as a reviewer but yea haha

phys2bids/bids_units.py
vinferrer: if you do ``orig_unit.lower() in unit aliases.keys()`` that could be problematic because we are expecting inputs that have both the unit and the multiplier. For example the unit ``msec`` that would trigger a false response to that statement whereas in this loop you are making sure that when ``msec`` unit and ``sec`` u_key  appear in the iteration, then the condition ``if u_key in unit`` will be satisfied regardless of the prefix

vinferrer: that's why we have to go through the hole dict. The condition you propouse does not satisfy this example which is going to be very common

smoia:
@vinferrer , if you checked that everything else is BIDS 1.3.0 compatible, does this close #185?

vinferrer:
I wouldn't say so. We would need to take a closer look

vinferrer:
In any case, do not merge this yet. I need to do the unit test so we don't loose code-coverage

eurunuela:
@vinferrer you still need another reviewer to approve in order to merge this PR in.

Anyway, better if you make a draft PR next time to avoid this being merged.

vinferrer:
yeah i know, is been while since i don't do a PR

smoia:
@eurunuela @vinferrer actually it's my fault. I thought the PR was not a draft any more and took it out. I've moved it back to draft again.
I know we didn't merge it yet but we should follow the practices in #197. One of those practice is to not approve a PR that hasn't been graduated from release, and in general not to approve it until it's ready for merge - meaning that title and labels have to be auto-friendly.

vinferrer:
I think this is ready for review now. I would like to comment that only when merging with the master branch there was a problem with test but it seems like it's solve for now. Trimming the testing files would be ideal to see if that's the problem

vinferrer:
So now we need one reviewer more to approve

phys2bids/tests/test_bids_units.py
```diff
@@ -0,0 +1,12 @@
...
+
+def test_bidsify_units():
+    # test normal unit
+    bidsify_units("V")
+    # test unit with standard prefix
+    bidsify_units("centik")
+    # test unit with not standard prefix
+    bidsify_units("matV")
+    # test unit that's not bids standard
+    bidsify_units("mmlie")
```
eurunuela, line 12: We should be checking that `bidsify_units( )` returns the correct units.

vinferrer: true, completely forgot. I just checked the coverage

vinferrer: done

smoia:
> I think this is ready for review now. I would like to comment that only when merging with the master branch there was a problem with test but it seems like it's solve for now. Trimming the testing files would be ideal to see if that's the problem

@vinferrer , if you have a look at the tests, the "failure" comes from codecov, that is alerting the loss of coverage. The CI seem to pass without problem! So it should be all ok with your new tests!

eurunuela:
I think this is ready unless @rmarkello has comments to make in his review.

phys2bids/bids_units.py
```diff
@@ -0,0 +1,94 @@
...
+            new_unit = unit[-len(u_key):]
+            # check that it is at the end
+            if new_unit == u_key:
```
rmarkello, line 76: Would something like:

```python
if unit.endswith(u_key):
   new_unit = unit[-len(u_key):]
```

be more parsimonious here?

phys2bids/bids_units.py
```diff
@@ -0,0 +1,94 @@
...
+                # leave only the prefix
+                unit = unit.replace(u_key, "")
```
rmarkello, line 78: I think that `unit = unit[:len(u_key)]` might be safer here. For example, if someone (for whatever reason) was using units of nano Newtons and coded it as `nn` then this would convert the entire thing to `''`.

phys2bids/bids_units.py
```diff
@@ -0,0 +1,94 @@
...
+                    # for every prefix alias
+                    for p_key in prefix_aliases.keys():
+                        # if the alias is the same
+                        if p_key == unit:
+                            # return the units bids corrected both for prefix and unit
+                            return prefix_aliases[p_key] + new_unit
+                    # return the unit bids corrected but not the prefix
+                    LGR.warning(f'The given unit prefix {unit} does not have aliases, '
+                                f'passing it as is')
+                    return unit + new_unit
```
rmarkello, line 90: Since all that's left _should_ be the prefix, you should just be able to try and get it from the dictionary (rather than iterating over the dictionary):

```python
return prefix_aliases.get(p_key, '') + new_unit
```

You can build in a check for if that `.get()` call returns `''` and raise the warning if you'd like, too.

phys2bids/bids_units.py
vinferrer: Yeah that's true but ``unit = unit[:len(u_key)]`` has also it's problems, for example, if the unit is 'v', when you use that expression the result would be that unit = 'v' again rendering useless the function more commonly. Maybe we can use this option but then we need to restructure the function. We would need to add an if for the cases where the unit does not have prefixes

phys2bids/bids_units.py
vinferrer: done

phys2bids/bids_units.py
vinferrer: done

phys2bids/bids_units.py
smoia: @vinferrer is it possible that @rmarkello was suggesting to do `unit = unit[-len(u_key):]`?
Would that solve both problems?

smoia: Sorry, similar mistake. `unit = unit[:-len(u_key)]` would solve it.
However, now that line 73 is different, line 76 is not necessary any more.
It could become (starting line 73):
`        if unit.endswith(u_key):
            new_unit = unit[-len(u_key):]
            # check that it is at the end
            unit = unit[:-len(u_key)]

            if unit != '':
                # for every prefix alias
                prefix = prefix_aliases.get(unit, '')
                if prefix == '':
                    LGR.warning(f'The given unit prefix {unit} does not have aliases, '
                                f'passing it as is')`

phys2bids/bids_units.py
```diff
@@ -0,0 +1,92 @@
...
+        BIDSified alias of input unit
+
+    Notes
+    -----
+    This function should implement a double check, one for unit and
+    the other for prefixes (e.g. "milli"). However, that is going to be tricky,
+    unless there is a weird way to multiply two dictionaries together.
+    """
+    # call prefix and unit dicts
+    new_unit = ""
```
smoia, line 68:
```suggestion
    new_unit = ''
```
I know it's _peccata minuta_, but since the rest of phy2bids (and the function) uses `'` for strings, it would be nice to maintain the same style.

phys2bids/tests/test_bids_units.py
```diff
@@ -0,0 +1,12 @@
+from phys2bids.bids_units import bidsify_units
+
+
+def test_bidsify_units():
+    # test normal unit
+    assert bidsify_units("V") == "v"
```
smoia, line 6:
```suggestion
    assert bidsify_units('V') == 'v'
```

Same as before. I know it's _peccata minuta_, but since the rest of phy2bids uses `'` for strings, it would be nice to maintain the same style.
Same for the following lines.

phys2bids/bids_units.py
```diff
@@ -0,0 +1,92 @@
...
+        unit = orig_unit.lower()
+        # check that u_key is part of unit
+        if unit.endswith(u_key):
+            new_unit = unit[-len(u_key):]
+            # check that it is at the end
+            if new_unit == u_key:
+                # leave only the prefix
+                unit = unit.replace(u_key, "")
+                # check prefix in not null
+                if unit != "":
```
smoia, line 80:
```suggestion
                if unit != '':
```
Same as before.

phys2bids/bids_units.py
vinferrer: I think i got it, could you check the lastest commit @rmarkello

phys2bids/tests/test_bids_units.py
```diff
@@ -0,0 +1,12 @@
+from phys2bids.bids_units import bidsify_units
+
+
+def test_bidsify_units():
```
smoia, line 4: I have a doubt:
The first two asserts should be failing.
In fact, they should be:
```
    assert bidsify_units('V') == 'V'
    # test unit with standard prefix
    assert bidsify_units('centik') == 'cK'
```

vinferrer: I know what you mean, let me check why they are not, actually i could just modify them

vinferrer: wait, no they shouldn't

vinferrer: they are actually in lowercase as that the way they come out of the function, remember?

vinferrer: we could make the function output the unit in upper case

phys2bids/bids_units.py
```diff
@@ -0,0 +1,92 @@
...
+    unless there is a weird way to multiply two dictionaries together.
+    """
+    # call prefix and unit dicts
+    new_unit = ''
+    # for every unit alias in the dict
+    for u_key in unit_aliases.keys():
+        unit = orig_unit.lower()
+        # check that u_key is part of unit
+        if unit.endswith(u_key):
+            new_unit = unit[-len(u_key):]
```
smoia, line 74:
```suggestion
            new_unit = unit_aliases[u_key]
```

phys2bids/bids_units.py
```diff
@@ -0,0 +1,92 @@
...
+            # check that it is at the end
+            if new_unit == u_key:
+                # leave only the prefix
+                if unit != u_key:
+                    unit = unit[:-len(u_key)]
+                    # for every prefix alias
+                    print(unit)
```
smoia, line 81: This can be removed
```suggestion

```

phys2bids/tests/test_bids_units.py
smoia: I think that changing the way `new_unit` is updated (see suggestion) should solve the problem.
The unit is not always in upper case (for instance, `mol`, `s`, or `Hz`)

phys2bids/bids_units.py
```diff
@@ -0,0 +1,92 @@
...
+        BIDSified alias of input unit
+
+    Notes
+    -----
+    This function should implement a double check, one for unit and
+    the other for prefixes (e.g. "milli"). However, that is going to be tricky,
+    unless there is a weird way to multiply two dictionaries together.
+    """
+    # call prefix and unit dicts
+    new_unit = ''
```
smoia, line 68: Now that I think about it, this can be removed too!
```suggestion

```

phys2bids/bids_units.py
smoia: With this suggestion and the `if` statement before, line 76 not only becomes unnecessary, but also wrong.
You can refactor line 73 to line 85 as:
```


        if unit.endswith(u_key):
            new_unit = unit_aliases[u_key]
            unit = unit[:-len(u_key)]
            if unit != '':
                # for every prefix alias
                prefix = prefix_aliases.get(unit, '')
                if prefix == '':
                    LGR.warning(f'The given unit prefix {unit} does not have aliases, '
                                f'passing it as is')
```

vinferrer: I think your logic here is wrong, but i will check

smoia: @vinferrer, your code is usually more solid than any of ours (while mine is the buggiest), so please check. However, in this case I'm not only passing the corrected test, but I'm also passing a couple of cases more (I'm trying to play with upper/lower cases in unit and/or prefixes).

I added a line to handle non-BIDS prefixes that are uppercase (and should stay uppercase if the idea is to return them unchanged), it's in [my branch](https://github.com/smoia/phys2bids/blob/enh/bids-units/phys2bids/bids_units.py) (lines 77-80), together with a couple of assertions more in the [test](https://github.com/smoia/phys2bids/blob/enh/bids-units/phys2bids/tests/test_bids_units.py). If you want, you can merge it this PR, or find a better solution!

phys2bids/bids_units.py
```diff
@@ -0,0 +1,92 @@
...
+    # for every unit alias in the dict
+    for u_key in unit_aliases.keys():
+        unit = orig_unit.lower()
```
rmarkello, line 71: I believe the `unit = orig_unit.lower()` can be outside the for loop, here.

phys2bids/bids_units.py
```diff
@@ -0,0 +1,92 @@
+import logging
+
+LGR = logging.getLogger(__name__)
+
+unit_aliases = {
```
rmarkello, line 5: We're going to want to re-order this dictionary such that there are fewer clashes. For example, if I pass:

``` python
>>> bidsify_unit('nanonetwon')
```

I get:

```
The given unit prefix nanonewto does not have aliases, passing it as is
'nanonewton'
```

because the `'n'` key is hit before the 'newton'`.

Similarly, if I pass:

```python
>>> bidsify_unit('1/hertz')
```

I get the output:

```
The given unit prefix 1/ does not have aliases, passing it as is
'1/hertz'
```

because `'hertz'` is checked before `'1/hertz'`. 

There aren't a TON of these clashes, but they will be important so we should either make the `bidsify_unit()` function a bit more robust to these cases or re-order the `unit_aliases` dictionary to attempt to avoid them. I _think_ the latter might be the easier thing to do, but maybe that's naive of me 🤷‍♂️

vinferrer: Okay, I reordered the dictionary and even more I added a loop to the test function that checks all the basic units work. This could be useful when adding new units in the dict. Do you think is an overkill?

phys2bids/bids_units.py
vinferrer: It cannot because unit gets modified in the loop therefore with every new iteration we need to restart it

rmarkello: What if you do:

```python
orig_unit = orig_unit.lower()
for u_key in unit_aliases.key():
    if orig_unit.endswith(u_key):
        new_unit = unit_aliases[u_key]
        unit = orig_unit[:-len(u_key)]
```

`str.lower()` is relatively cheap, but no reason not to do small optimizations! Let me know if you think that will work or if I'm overlooking something...

vinferrer:
@rmarkello I think we are done with this, unless someone has any comments?

PR #223, opened 2020-05-28
Fix azure pipelines non-failure bug

### LINKED ISSUE #215: Azure tests pass while the actual test code raises an error
## Expected Behavior

Azure tests should fail when pytest raises an error.

## Actual Behavior

As discussed in #206, the Azure tests pass even if the pytest code raises an error.

## Possible solution

As @rmarkello said in #206 : 

> I think adding `failOnStderr: true` to the "Run tests" script in the `.azure-pipelines/windows.yml` definition file should catch this.

rmarkello:
Closes #215 

## Proposed Changes

- Adds `failOnStderr` entry to YAML for running the tests to, hopefully, fail when there are errors

Not sure if this will _actually_ fix things given that they seem to magically be working again, but will see how things go with this PR. May purposefully break something to see if I can reproduce the errors we were seeing before, but will let this run through once, first!

### DIFF
--- a/.azure-pipelines/windows.yml
+++ b/.azure-pipelines/windows.yml
@@ -36,13 +36,11 @@ jobs:
         mkdir for_testing
         cd for_testing
         cp ../setup.cfg .
-        python -m pytest --doctest-modules --cov-report term-missing --cov=phys2bids --pyargs phys2bids
-        cp ./.coverage ../
+        python -m pytest --doctest-modules --cov-report term-missing --cov-report xml --cov=phys2bids --pyargs phys2bids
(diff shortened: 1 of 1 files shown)


rmarkello:
HAH! Got it 💯 That was _really_ annoying (mostly Azure's fault), but this should be working now.

@eurunuela, if you wouldn't mind doing the honors of a review 😄

eurunuela:
Given that this is a bug fix PR, I'll ask @smoia to review this before merging.

smoia:
Sure! Approved!

PR #231, opened 2020-06-11
Update reference API

rmarkello:
Closes #??

## Proposed Changes

  - Adds info to reference API

What other functions /  classes should be added to the API? Happy to add whatever but I don't think _everything_ needs to be on it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,4 +124,5 @@ dmypy.json
 # Pyre type checker
 .pyre/
 
+docs/generated/
 .vscode/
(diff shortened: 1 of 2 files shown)


smoia:
I didn't try to build it locally yet, but I think it's ok!

I wonder if it makes sense to add a heuristic file as well?

What do you think?

rmarkello:
Like, a template heuristic file? Or the function for processing the heuristic file (i.e., `use_heur()` or whatever it's called)?

In my mind the API should just be documenting functions in the codebase, so if you were suggesting the former I'd err on the side of no. But I can add the latter, if you'd like?

smoia:
Mmm, yeah, you're right. I don't think it's necessary to document the latter though - I think there's everything we need!

PR #234, opened 2020-06-13
Move bids-related functions to dedicated file

smoia:
Closes #

## Proposed Changes

  - Rename `bids_units.py` as `bids.py`
  - Move `use_heuristic` from `phy2bids.py` to `bids.py`
  - Adapt tests consequently.


smoia:
This is a way to split a bit more #206 and merge some bits before cutting a release for OHBM.

PR #226, opened 2020-06-04
Update testing to use trimmed files

eurunuela:
Closes #221

## Proposed Changes

  - Testing uses trimmed files to reduce computation time.

PR #241, opened 2020-06-15
Update testing to use short files.

### LINKED ISSUE #221: CI tests failing due to bad data download
## Summary

It seems that since a couple of days, both Azure and Travis are failing some or all the jobs due to a "bad" retrieval of the testing data.

The typical error is either:
```
urllib.error.ContentTooShortError: <urlopen error retrieval incomplete: got only 17361695 out of 17378311 bytes>

urlopen error retrieval incomplete: got only 17373909 out of 17378311 bytes
```
or 
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

All those issues started more or less after OSF latest [system maintenance](https://status.cos.io/), but I can't figure out why they happen.

The only reason I can think about could be that we're opening too many connections to OSF to retrieve the data.
In that case, we can do one or more of the following:
  - Momentarily turn off Azure. It's not the best solution, but since it's not working properly anyway (see #215), that may let Travis do its job and at least we would have one testing suite working. 
  - [Cache the tests files in Travis](https://docs.travis-ci.com/user/caching/#arbitrary-directories). This won't speed things up, but maybe it will help. I don't know if we can do it in Azure.
  - Change host site for the files. For instance, we can open a google account for physiopy and put some test files in drive.
  - Configure the tests to retrieve files only after a certain time passed since the last retrieval. This will slow down the tests though.
  - ...

Any suggestion?

I read that using request instead of urllib could also be a possibility, I wonder if it would solve the issue.

**In the meatime, if you're reviewing a PR, please check it out and test it locally,** running flake8, pytest, and sphinx.

## Additional detail

Check any recent Travis or Azure log

## Next Steps

 - Figure out where the issue comes from.
 - Figure out what to do.

eurunuela:
Co-authored-by: Rachael Stickland <50215726+RayStick@users.noreply.github.com>

Closes #221 and fixes #226.

Proposed Changes:

- Testing uses trimmed files to reduce computation time.

### DIFF
--- a/phys2bids/tests/test_integration.py
+++ b/phys2bids/tests/test_integration.py
@@ -225,7 +225,7 @@ def test_integration_multifreq(multifreq_lab_file):
     assert json_data['Columns'] == ['CO2']
 
     """
-    Checks 100 Hz output
+    Checks 500 Hz output
     """
     # Read log file of frequency 625 (note that this file is not the logger file)
     with open(os.path.join(test_path, 'Test1_multifreq_onescan_500.0.log')) as log_info:
--- a/phys2bids/tests/test_integration.py
+++ b/phys2bids/tests/test_integration.py
@@ -21,23 +21,22 @@ def check_string(str_container, str_to_find, str_expected, is_num=True):
         return str_expected in str_found
 
 
-def test_logger():
+def test_logger(multifreq_lab_file):
     """
     Tests the logger
     """
 
-    # Move into folder
-    test_path = resource_filename('phys2bids', 'tests/data')
-    subprocess.run(f'cd {test_path}', shell=True, check=True)
+    test_path, test_filename = os.path.split(multifreq_lab_file)
 
     # Input arguments
-    test_filename = 'tutorial_file.txt'
-    test_chtrig = 1
+    test_chtrig = 3
+    test_ntp = 1
     test_outdir = test_path
 
     # Phys2bids call through terminal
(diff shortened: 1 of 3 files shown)


eurunuela:
@rmarkello @vinferrer I'm sorry for what happened with #226 . This should be the good one.

phys2bids/tests/test_integration.py
```diff
@@ -161,66 +158,125 @@ def test_integration_multifreq(multifreq_acq_file):
...
-    assert math.isclose(json_data['SamplingFrequency'], 10000.0)
-    assert math.isclose(json_data['StartTime'], 10.4251)
-    assert json_data['Columns'] == ['time', 'RESP - RSP100C',
-                                    'MR TRIGGER - Custom, HLT100C - A 5', 'PPG100C', 'CO2', 'O2']
+    assert math.isclose(json_data['SamplingFrequency'], 100.0)
+    assert math.isclose(json_data['StartTime'], -0.3057)
+    assert json_data['Columns'] == ['CO2']
+
+    """
+    Checks 100 Hz output
```
rmarkello, line 228: Is this next section checking the 100 Hz or 500 Hz output?

eurunuela: Thanks @rmarkello ! Copy-pasting typo 😅

eurunuela:
Thank you so much @rmarkello @vinferrer for the quick review. Thank you @RayStick for helping with the files!!

Issue #199, opened 2020-04-02
Integration with BIDScoin


marcelzwiers:
I am the author of BIDScoin, a DICOM to BIDS converter:
https://github.com/Donders-Institute/bidscoin

# Goal 
BIDScoin doesn't convert physiological data but supports plugins and I was wondering how feasible it would be to write a plugin-wrapper around `phys2bids` tool(s) so that it integrates directly with BIDScoin.

#How would that work and what would be the benefit?
BIDScoin would pass the heuristics (e.g. BIDS-labels) it extracted from the imaging source data to `phys2bids`, which would otherwise not be available so easily. It would allow to convert both the imaging and physiological data in one pass, making the BIDS output data more consistent. All this is a bit of a long shot from my side, since BIDScoin I don't know much about the working and design of `phys2bids`, I'm just exploring to see if there is common ground here and I would be very interested to hear your thoughts on this...

Cheers,
Marcel

smoia:
Hello @marcelzwiers , thank you for opening this issue!
I think it could be feasible to write a plugin-wrapper around `phys2bids` to integrate the two programs together. Remember, however, that for the moment `phys2bids` does not support physiological file recorded with the MRI, but only with external instruments such as Biopac and ADInstruments. This doesn't mean we won't support them in the future, but for the moment other packages that do so exist already.

This said, the main challenge I foresee would be associating the right DICOM files to the right physiological recordings. That done, passing the right labels to the main function of phy2bids should be quite easy - especially if we refactor our code as planned in #44 to have heuristic files that are a list of `if ..elif` statements.

We have a developer call next Thursday, as tracked in #194. Could you/would you join? We could talk about it all together.

marcelzwiers:
For follow up, see:
- https://github.com/ohbm/hackathon2021/issues/12
- https://github.com/Donders-Institute/bidscoin/issues/77 (to some extent)

PR #244, opened 2020-06-16
Generate participants.tsv if it doesn't exist or update it if subject is missing in the file

### LINKED ISSUE #242: Automatically generate BIDS compulsory agnostic files and update them if they already exist
comes from #185. When a heuristic file is provided and data is bidsdifyed agnostic files (README, participants.tsv and dataset_description.json) should be generated
## Detailed Description
When a heuristic file is provided and data is bidsdifyed agnostic files should be generated:

- [ ] dataset_description.json
       Fields
       - Name (Required)
       - BIDSVersion (Required)
       - DatasetType (Recommended)
       - License (Recommended)

- [ ] Participants.tsv
       Fields
       - participant_id (Required)
       - age (Recommended)
       - sex (Recommended)
       - handedness (Recommended)

- [ ] README

## Context / Motivation
comply BIDS 1.4 standard

## Possible Implementation
Create one function per file where we check if the file exists in the dataset dir, in case it doesn't generate it and cover at least the required fields, preferably also the  recommended  and add them to the pipeline.

eurunuela:
Partly closes #242 and #185 

## Proposed Changes

  - Generate participants.tsv if it doesn't exist already.

### DIFF
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -1,6 +1,8 @@
+from csv import reader
 import os
 import logging
 
+import yaml
 from pathlib import Path
 
 from phys2bids import utils
@@ -170,3 +172,58 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
     heurpath = os.path.join(fldr, f'{name}physio')
 
     return heurpath
+
+
+def participants_file(indir, outdir, yml, sub):
+    """[summary]
+
+    Args:
+        indir ([type]): [description]
+        outdir ([type]): [description]
+        yml ([type]): [description]
+        sub ([type]): [description]
+    """
+
+    file_path = os.path.join(outdir, 'participants.tsv')
+    if not os.path.exists(file_path):
+        # Read yaml info if file exists
+        if os.path.exists(os.path.join(indir, yml)):
+            with open(os.path.join(indir, yml)) as f:
+                yaml_data = yaml.load(f, Loader=yaml.FullLoader)
+            p_id = yaml_data['participant']['participant_id']
+            p_age = yaml_data['participant']['age']
+            p_sex = yaml_data['participant']['sex']
+            p_handedness = yaml_data['participant']['handedness']
+        else:
+            # Fill in with data from phys2bids
+            p_id = sub
+            p_age = 'n/a'
+            p_sex = 'n/a'
+            p_handedness = 'n/a'
+
+        # Write to participants.tsv file
+        header = ['participant_id', 'age', 'sex', 'handedness']
+        utils.append_list_as_row(file_path, header)
+
+        participants_data = [p_id, p_age, p_sex, p_handedness]
+        utils.append_list_as_row(file_path, participants_data)
+
+    else:  # If participants.tsv exists only update when subject is not there
+        # Find participant_id column in header
+        pf = open(file_path, 'r')
+        header = pf.readline().split("\t")
+        pf.close()
+        p_id_idx = header.index('participant_id')
+
+        # Check if subject is already in the file
+        sub_exists = False
+        with open(file_path) as pf:
+            tsvreader = reader(pf, delimiter="\t")
+            for line in tsvreader:
+                if sub in line[p_id_idx]:
+                    sub_exists = True
+                    break
+        # Only append to file if subject is not in the file
+        if not sub_exists:
+            participants_data = [sub, 'n/a', 'n/a', 'n/a']
+            utils.append_list_as_row(file_path, participants_data)
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -195,7 +195,7 @@ def participants_file(outdir, yml, sub):
     if not os.path.exists(file_path):
         LGR.warning('phys2bids could not find participants.tsv')
         # Read yaml info if file exists
-        if '.yml' in yml and os.path.exists(yml):
+        if '.yml' in yml is not None and os.path.exists(yml):
             LGR.info('Using yaml data to populate participants.tsv')
             with open(yml) as f:
                 yaml_data = yaml.load(f, Loader=yaml.FullLoader)
@@ -239,6 +239,7 @@ def participants_file(outdir, yml, sub):
         # Only append to file if subject is not in the file
         if not sub_exists:
             LGR.info(f'Appending subjet sub-{sub} to participants.tsv ...')
-            participants_data = 'n/a' * header_length
+            breakpoint()
+            participants_data = ['n/a'] * header_length
             participants_data[p_id_idx] = f'sub-{sub}'
             utils.append_list_as_row(file_path, participants_data)
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -239,7 +239,6 @@ def participants_file(outdir, yml, sub):
         # Only append to file if subject is not in the file
         if not sub_exists:
             LGR.info(f'Appending subjet sub-{sub} to participants.tsv ...')
-            breakpoint()
             participants_data = ['n/a'] * header_length
             participants_data[p_id_idx] = f'sub-{sub}'
             utils.append_list_as_row(file_path, participants_data)
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -183,18 +183,22 @@ def participants_file(indir, outdir, yml, sub):
         yml ([type]): [description]
         sub ([type]): [description]
     """
-
+    LGR.info('Updating participants.tsv ...')
     file_path = os.path.join(outdir, 'participants.tsv')
     if not os.path.exists(file_path):
+        LGR.warning('phys2bids could not find participants.tsv')
         # Read yaml info if file exists
         if os.path.exists(os.path.join(indir, yml)):
+            LGR.info('Using yaml data to populate participants.tsv')
             with open(os.path.join(indir, yml)) as f:
                 yaml_data = yaml.load(f, Loader=yaml.FullLoader)
-            p_id = yaml_data['participant']['participant_id']
+            p_id = sub
             p_age = yaml_data['participant']['age']
             p_sex = yaml_data['participant']['sex']
             p_handedness = yaml_data['participant']['handedness']
         else:
+            LGR.info('No yaml file was provided. Using phys2bids data to '
+                     'populate participants.tsv')
             # Fill in with data from phys2bids
             p_id = sub
             p_age = 'n/a'
@@ -209,6 +213,7 @@ def participants_file(indir, outdir, yml, sub):
         utils.append_list_as_row(file_path, participants_data)
 
     else:  # If participants.tsv exists only update when subject is not there
+        LGR.info('phys2bids found participants.tsv. Updating if needed...')
         # Find participant_id column in header
         pf = open(file_path, 'r')
         header = pf.readline().split("\t")
@@ -225,5 +230,6 @@ def participants_file(indir, outdir, yml, sub):
                     break
         # Only append to file if subject is not in the file
         if not sub_exists:
+            LGR.info(f'Appending subjet {sub} to participants.tsv ...')
             participants_data = [sub, 'n/a', 'n/a', 'n/a']
             utils.append_list_as_row(file_path, participants_data)
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -1,9 +1,9 @@
-from csv import reader
 import os
 import logging
+from csv import reader
+from pathlib import Path
 
 import yaml
-from pathlib import Path
 
 from phys2bids import utils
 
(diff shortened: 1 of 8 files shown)


eurunuela:
@smoia @vinferrer @tsalo what should we do when `participants.tsv` already exists?

1. Warn the user and not update the file.
2. Warn the user and update the file with the yaml content.
3. Have an overwrite option in the parser and update the file.

tsalo:
For the participants.tsv, I think that updating an existing file will be the norm after converting the first participant, so I vote 2, except when the participant already exists in the file.

smoia:
Totally agree with @tsalo on this.
Check if the file exists,
    if so check if the subject id is present,
        if so update with missing information (but without overwriting values),
        otherwise add subject id (and values) to the list
Then update the file in the folder or write it.

I think we can follow this steps for all the files.

phys2bids/utils.py
```diff
@@ -280,3 +282,54 @@ def load_heuristic(heuristic):
...
+def append_list_as_row(file_name, list_of_elem):
+    # Open file in append mode
+    with open(file_name, 'a+', newline='') as write_obj:
+        # Create a writer object from csv module
+        csv_writer = writer(write_obj, delimiter='\t')
+        # Add contents of list as last row in the csv file
+        csv_writer.writerow(list_of_elem)
+
+
+def participants_file(indir, outdir, yml, sub):
```
vinferrer, line 296: I think this function should go in the bids.py script since is part of the bidsdifiy process

eurunuela: Done. Thanks! I'll push those changes in a minute, once I fix the yaml dependency issue.

eurunuela:
Tests pass but we're losing coverage. I'll add an extra call to phys2bids in the integration test to cover part of the new function. Given that this PR is critical for OHBM, I think we can add the unit tests later. 

What do you think @smoia @vinferrer ?

eurunuela:
Adding the extra call to phys2bids makes the test much slower... How about we add the `dataset_description.json` functionality first and then we work on the unit tests to avoid tests taking this long.

vinferrer:
> 
> 
> Adding the extra call to phys2bids makes the test much slower... How about we add the `dataset_description.json` functionality first and then we work on the unit tests to avoid tests taking this long.
Sure, Do you wanna open an issue so we don't forget about it?

vinferrer:
> 
> 
> > Adding the extra call to phys2bids makes the test much slower... How about we add the `dataset_description.json` functionality first and then we work on the unit tests to avoid tests taking this long.
> > 
Sure, Do you wanna open an issue so we don't forget about it?

phys2bids/bids.py
```diff
@@ -170,3 +172,58 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
...
+        sub ([type]): [description]
+    """
+
+    file_path = os.path.join(outdir, 'participants.tsv')
+    if not os.path.exists(file_path):
+        # Read yaml info if file exists
+        if os.path.exists(os.path.join(indir, yml)):
+            with open(os.path.join(indir, yml)) as f:
+                yaml_data = yaml.load(f, Loader=yaml.FullLoader)
+            p_id = yaml_data['participant']['participant_id']
```
vinferrer, line 193: I think that in bids standard ``participant_id`` and sub are always the same therefore ``p_id=sub`` should be constant since the name of the subject has to be entered in the console argument -sub

vinferrer: In my opinion ```p_id=sub`` should be out of this condition

phys2bids/cli/run.py
```diff
@@ -125,6 +125,11 @@ def _get_parser():
                           type=str,
                           help='full path to store channels plot ',
                           default='')
+    optional.add_argument('-yml', '--participant-yml',
+                          dest='yml',
+                          type=str,
+                          help='file with info needed to generate participant.tsv file ',
+                          default='participants.yml')
```
vinferrer, line 132: Do we need a default? in that case, shouldn't this default be  ``phys2bids/heuristics/participant.yml``?

phys2bids/phys2bids.py
```diff
@@ -296,6 +297,12 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
                       phys_in.num_timepoints_found, uniq_freq,
                       phys_out[uniq_freq].start_time, outfile)
 
+    # Generate participants.tsv file if it doesn't exist already.
+    # Update the file if the subject is not in the file.
+    # Do not update if the subject is already in the file.
+    if heur_file:
+        participants_file(indir, outdir, yml, sub)
```
vinferrer, line 304: In order to save lines of code, this should go in line 282, we avoid the extra if and make sure that sub is given by the user

phys2bids/bids.py
eurunuela: Fair enough. So should we use `sub` instead of the yaml data?

phys2bids/cli/run.py
eurunuela: I wrote `participants.yml` because the tsv file is `participants.tsv`. I know the one I sent you was `participant.yml`.

phys2bids/phys2bids.py
eurunuela: The problem there is we would be calling the function many times inside the for loop. But yeah, I'll add the `sub` check.

phys2bids/bids.py
vinferrer: Yeah that is what i was suggesting, since -sub has to be introduced for bidsdifying data

phys2bids/cli/run.py
vinferrer: Actually i was saying that maybe we don't need a default, since the participants function has an option when the file does not exist
```suggestion
                          default='l')
```

phys2bids/bids.py
```diff
@@ -170,3 +172,58 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
...
+            p_age = yaml_data['participant']['age']
+            p_sex = yaml_data['participant']['sex']
+            p_handedness = yaml_data['participant']['handedness']
+        else:
+            # Fill in with data from phys2bids
```
vinferrer, line 198: I think a warning here is appropiate since maybe the user is not aware of the ``participants.yml``

phys2bids/phys2bids.py
vinferrer: you could call it in line 267

phys2bids/cli/run.py
eurunuela: Done! Thanks for your suggestion!

eurunuela:
@vinferrer I think I've covered all your comments, let me know!

phys2bids/phys2bids.py
```diff
@@ -296,6 +297,12 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
                       phys_in.num_timepoints_found, uniq_freq,
                       phys_out[uniq_freq].start_time, outfile)
 
+    # Generate participants.tsv file if it doesn't exist already.
+    # Update the file if the subject is not in the file.
+    # Do not update if the subject is already in the file.
+    if heur_file and sub:
```
vinferrer, line 303: I am gonna insist, there is a if statement that is identical to this one in line 266, let's not duplicate code. I know you wanted to put the participants.tsv code at the end of the function but why not there? it doesn't influence the rest of the code

eurunuela: As I previously said, line 266 is part of a for loop. I don't think it makes sense to run `participants_file(indir, outdir, yml, sub)` for every frequency in the file.

phys2bids/phys2bids.py
eurunuela: Didn't see this comment. Sure, I'll move it up there.

phys2bids/phys2bids.py
eurunuela: Resolving this after seeing your comment above.

phys2bids/phys2bids.py
eurunuela: Okay, last commit took care of this. Resolving this comment now.

phys2bids/bids.py
```diff
@@ -170,3 +172,64 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
     heurpath = os.path.join(fldr, f'{name}physio')
 
     return heurpath
+
+
+def participants_file(indir, outdir, yml, sub):
+    """[summary]
```
smoia, line 178: Can you write the docstring please?

phys2bids/bids.py
```diff
@@ -1,6 +1,8 @@
+from csv import reader
 import os
 import logging
 
+import yaml
```
smoia, line 5: I _think_ that we decided all the "from x import y" to go at the end of their relative section (so "from csv import reader" should go before pathlib but after logging).
yaml is a non-core, so it goes after pathlib.

phys2bids/bids.py
```diff
@@ -170,3 +172,64 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
...
+                yaml_data = yaml.load(f, Loader=yaml.FullLoader)
+            p_id = sub
+            p_age = yaml_data['participant']['age']
+            p_sex = yaml_data['participant']['sex']
+            p_handedness = yaml_data['participant']['handedness']
+        else:
+            LGR.info('No yaml file was provided. Using phys2bids data to '
+                     'populate participants.tsv')
+            # Fill in with data from phys2bids
+            p_id = sub
```
smoia, line 203: Can you quickly check whether this should be `f'sub-{sub}'` or only `sub`

phys2bids/bids.py
```diff
@@ -170,3 +172,64 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
...
+            p_age = 'n/a'
```
smoia, line 204: These three columns are not mandatory. I would skip the addition of 'n/a' columns.

phys2bids/bids.py
```diff
@@ -170,3 +172,64 @@ def use_heuristic(heur_file, sub, ses, filename, outdir, record_label=''):
...
+        with open(file_path) as pf:
+            tsvreader = reader(pf, delimiter="\t")
+            for line in tsvreader:
+                if sub in line[p_id_idx]:
+                    sub_exists = True
+                    break
+        # Only append to file if subject is not in the file
+        if not sub_exists:
+            LGR.info(f'Appending subjet {sub} to participants.tsv ...')
+            participants_data = [sub, 'n/a', 'n/a', 'n/a']
```
smoia, line 234: What if the `participants.tsv` file didn't have four columns, but say, only 1 or 2?
I would add a check here for the number of 'n/a' to add.

phys2bids/cli/run.py
```diff
@@ -125,6 +125,11 @@ def _get_parser():
                           type=str,
                           help='full path to store channels plot ',
                           default='')
+    optional.add_argument('-yml', '--participant-yml',
```
smoia, line 128: When we're going to work with #199, -yml might become confusing (since the heuristic will be a yaml file). Could you find a different short argument?

phys2bids/phys2bids.py
```diff
@@ -264,6 +265,10 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
 
     if heur_file and sub:
         LGR.info(f'Preparing BIDS output using {heur_file}')
+        # Generate participants.tsv file if it doesn't exist already.
+        # Update the file if the subject is not in the file.
+        # Do not update if the subject is already in the file.
+        participants_file(indir, outdir, yml, sub)
```
smoia, line 271: What if we remove `indir` and just ask for the absolute path to the yaml file?

smoia: In fact, in order to simplify the arguments of the CLI, we could remove `indir` for good...

phys2bids/bids.py
vinferrer: I disagree, these are the Recommended options, which are not mandatory but are in a higher level than the optional ones

phys2bids/bids.py
eurunuela: Thanks! I forgot I had to do that!

phys2bids/bids.py
eurunuela: `sub` only returns the number. I guess it's got to be `f'sub-{sub}'`

phys2bids/phys2bids.py
eurunuela: Sure! Sounds good.

eurunuela: We can do that in another PR.

phys2bids/bids.py
eurunuela: I don't agree with this. As we discussed yesterday (I think with @vinferrer and @tsalo), phys2bids should recommend those as BIDS does.

phys2bids/bids.py
eurunuela: Good point!

eurunuela:
@smoia I think I've covered the most important comments. Please let me know.

eurunuela:
There's still a bug on that previous commit. Fix coming.

phys2bids/bids.py
smoia: But we're not recommending. We're just adding empty columns.

If we want to recommend, then let's add a warning that states "Hey you should add those columns by hand!"

phys2bids/cli/run.py
smoia: What about calling it -ptable or similar?

eurunuela: The description data should also go in the same yaml file to make the user's life easier... so I'm not sure `-ptable` should be its name. When we work with #199, will the heuristic be always a yaml? If that's the case, then `-heur` will be free to use.

phys2bids/bids.py
eurunuela: I have no strong opinions about it but would like to know what @vinferrer and @tsalo think

phys2bids/cli/run.py
smoia: The file will be given through `-heur`, but my concern is that users might get confused by `-yml`.
Anyway, let's keep it like this.

PR #245, opened 2020-06-16
Create bids_ignore directory and redirect channels plot, trigger plot, logger output and subject log file to this directory

vinferrer:
## Proposed Changes

  - Create extra folder under the output dir
  - Eliminate channels plot path output option and output channel plot always here, channel plot not optional anymore
  - Eliminate trigger plot path output option and output channel plot always here, channel plot not optional anymore
  - Change logger file output folder to extra
  - Same with subject log file

### DIFF
--- a/phys2bids/phys2bids.py
+++ b/phys2bids/phys2bids.py
@@ -207,7 +207,7 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
         # #!# Get option of no trigger! (which is wrong practice or Respiract)
         phys_in.check_trigger_amount(thr, num_timepoints_expected, tr)
         LGR.info('Plot trigger')
-        plot_path = os.path.join(outdir,
+        plot_path = os.path.join(extra_dir,
                                  os.path.splitext(os.path.basename(filename))[0])
         if sub:
             plot_path += f'_sub-{sub}'
--- a/phys2bids/phys2bids.py
+++ b/phys2bids/phys2bids.py
@@ -133,12 +133,13 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     outdir = utils.check_input_dir(outdir)
     utils.path_exists_or_make_it(outdir)
     # generate extra path
-    utils.path_exists_or_make_it(outdir + "/extra")
+    extra = outdir + "/extra"
+    utils.path_exists_or_make_it(extra)
     # Create logfile name
     basename = 'phys2bids_'
     extension = 'tsv'
     isotime = datetime.datetime.now().strftime('%Y-%m-%dT%H%M%S')
-    logname = os.path.join(outdir, (basename + isotime + '.' + extension))
+    logname = os.path.join(extra, (basename + isotime + '.' + extension))
 
     # Set logging format
     log_formatter = logging.Formatter(
--- a/phys2bids/phys2bids.py
+++ b/phys2bids/phys2bids.py
@@ -133,7 +133,7 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     outdir = utils.check_input_dir(outdir)
     utils.path_exists_or_make_it(outdir)
     # generate extra path
-    extra_dir = outdir + "/extra"
+    extra_dir = outdir + "/bids_ignore"
     utils.path_exists_or_make_it(extra_dir)
     # Create logfile name
     basename = 'phys2bids_'
--- a/phys2bids/phys2bids.py
+++ b/phys2bids/phys2bids.py
@@ -133,13 +133,13 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     outdir = utils.check_input_dir(outdir)
     utils.path_exists_or_make_it(outdir)
     # generate extra path
-    extra = outdir + "/extra"
-    utils.path_exists_or_make_it(extra)
+    extra_dir = outdir + "/extra"
+    utils.path_exists_or_make_it(extra_dir)
     # Create logfile name
     basename = 'phys2bids_'
     extension = 'tsv'
     isotime = datetime.datetime.now().strftime('%Y-%m-%dT%H%M%S')
-    logname = os.path.join(extra, (basename + isotime + '.' + extension))
+    logname = os.path.join(extra_dir, (basename + isotime + '.' + extension))
 
     # Set logging format
     log_formatter = logging.Formatter(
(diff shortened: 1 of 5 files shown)


vinferrer:
Eneko and I were thinking that we should always direct the channels plot, trigger plot, logger output and subject log file. Therefore I am gonna eliminate user output options for channels and trigger plot. Any concerns?

smoia:
By "direct" you mean output?
Yes, it can be done. Once per input file rather than one per output.

eurunuela:
@vinferrer please make it ready for review if it actually is ready for review 😄

vinferrer:
i was wating for travis

vinferrer:
There seems to be a problem in the azure test:
``Could not get the latest source version for repository physiopy/phys2bids hosted on https://github.com/ using ref 0d4ed8d325a575a3534a09c8510af13d242b3a00. GitHub reported the error, "An error occurred while sending the request."``
Can we rerun it?

eurunuela:
I think @tsalo mentioned the Azure issue yesterday. Let me see if I have the same issue with my PR and I'll get back to you.

eurunuela:
@vinferrer why don't you make an empty commit to run the tests again? It's very weird that your test ran for less than 1 second in Azure. It's running on #244 and the error @tsalo mentioned was CodeCov related I think.

vinferrer:
Now the azure tests are not triggering

phys2bids/phys2bids.py
```diff
@@ -132,12 +132,14 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     # #!# This can probably be done while parsing?
     outdir = utils.check_input_dir(outdir)
     utils.path_exists_or_make_it(outdir)
-
+    # generate extra path
+    extra_dir = outdir + "/extra"
```
eurunuela, line 136:
```suggestion
    extra_dir = outdir + "/extras"
```

eurunuela: I'm not sure if this should be called `extras`, `others`, `phys2bids_others`...

What do you think @smoia ?

vinferrer: I don't mind changing the name, but let's decide and then i will do the hole change, because it is also in the integration tests

smoia: `phys2bids_logs`? `bids_ignore`? `extras`?

eurunuela: It's not just the logs in the folder so I'd avoid the word "logs". `bids_ignore` sounds good to me.

vinferrer: perfect

phys2bids/tests/test_integration.py
```diff
@@ -32,14 +33,14 @@ def test_logger(multifreq_lab_file):
     test_chtrig = 3
     test_ntp = 1
     test_outdir = test_path
-
+    extra_dir = test_path + '/bids_ignore/'
```
smoia, line 36: I'm writing it here, but can you check it in across tests?

As a general thing, you could make paths with `os.path.join(test_path, 'bids_ignore')` rather that with string unions.
It's better for compatibility since it doesn't assume `/` to be the separator (although probably azure would not break due to the same function being used later).

If you're using frequently this variable with the same initialisation you could make it a constant (just add it after the `import` section in capital letter). Then you can import it to other testing files (or from other testing files to this one).

smoia: (PS: Thank you for the `'`!)

phys2bids/phys2bids.py
```diff
@@ -132,12 +132,14 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     # #!# This can probably be done while parsing?
     outdir = utils.check_input_dir(outdir)
     utils.path_exists_or_make_it(outdir)
-
+    # generate extra path
+    extra_dir = outdir + '/bids_ignore'
```
eurunuela, line 136:
```suggestion
    extra_dir = os.path.join(outdir, 'bids_ignore')
```

phys2bids/tests/test_integration.py
```diff
@@ -368,10 +388,10 @@ def test_integration_info(samefreq_short_txt_file):
     subprocess.run(command_str, shell=True, check=True)
 
     # Check that plot all file is generated
-    assert os.path.isfile('Test_belt_pulse_samefreq_short.png')
+    assert os.path.isfile(os.path.join(test_outdir, 'bids_ignore/Test_belt_pulse_samefreq_short.png'))
```
smoia, line 391: Same concept here: since `os.path.join()` accepts more than two arguments, you can express this as:
    `assert os.path.isfile(os.path.join(test_outdir, 'bids_ignore','Test_belt_pulse_samefreq_short.png'))`

phys2bids/tests/test_integration.py
```diff
@@ -104,16 +111,21 @@ def test_integration_acq(samefreq_full_acq_file):
 
     test_path, test_filename = os.path.split(samefreq_full_acq_file)
     test_chtrig = 3
+    extra_dir = test_path + '/bids_ignore/'
```
eurunuela, line 114:
```suggestion
    extra_dir = os.path.join(test_path, 'bids_ignore')
```

phys2bids/tests/test_integration.py
```diff
@@ -151,30 +164,34 @@ def test_integration_multifreq(multifreq_lab_file):
 
     test_path, test_filename = os.path.split(multifreq_lab_file)
     test_chtrig = 3
+    extra_dir = test_path + '/bids_ignore/'
```
eurunuela, line 167:
```suggestion
    extra_dir = os.path.join(test_path, 'bids_ignore')
```

phys2bids/tests/test_integration.py
```diff
@@ -356,7 +376,7 @@ def test_integration_info(samefreq_short_txt_file):
     test_ntp = 158
     test_tr = 1.2
     test_thr = 0.735
-
+    extra_dir = test_path + '/bids_ignore/'
```
smoia, line 379: Same concept

phys2bids/tests/test_integration.py
```diff
@@ -294,6 +312,7 @@ def test_integration_heuristic(samefreq_short_txt_file):
     test_full_path = os.path.join(test_path, test_filename)
     test_chtrig = 1
     test_outdir = test_path
+    extra_dir = test_path + '/bids_ignore/'
```
eurunuela, line 315:
```suggestion
    extra_dir = os.path.join(test_path, 'bids_ignore')
```

phys2bids/tests/test_integration.py
eurunuela, line 379:
```suggestion
    extra_dir = os.path.join(test_path, 'bids_ignore')
```

phys2bids/tests/test_integration.py
eurunuela: Looks like we both asked for the same changes haha

smoia:
@vinferrer I just solved a conflict - pull from your remote before starting again!

vinferrer:
Okay I Think this is good to go

PR #243, opened 2020-06-16
Generate report with phys2bids outputs

eurunuela:
Closes #131 

## Proposed Changes

  - Adds the functionality to generate a report with phys2bids outputs on every call to phys2bids.

PR #255, opened 2020-06-18
Generate dataset_description.json and README.md when they do not exist and update the json file when the subject is missing

vinferrer:
Closes #243 

## Proposed Changes

  - Generate dataset_description.json if it doesn't exist already.
  - Generate README if it doesn't exist already.

### DIFF
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -278,8 +278,8 @@ def readme_file(outdir):
 
     """
     file_path = os.path.join(outdir, 'README.md')
-    text = 'Empty README, please fill in describing the dataset in more detail.'
     if not os.path.exists(file_path):
+        text = 'Empty README, please fill in describing the dataset in more detail.'
         LGR.warning('phys2bids could not find README,'
                     'generating it EMPTY, please fill in the necessary info')
         utils.writefile(file_path, '', text)
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -244,7 +244,7 @@ def participants_file(outdir, yml, sub):
             utils.append_list_as_row(file_path, participants_data)
 
 
-def data_description_file(outdir):
+def dataset_description_file(outdir):
     """
     Create dataset_description.json file if it does not exist.
     If it exists, do nothing.
@@ -265,3 +265,21 @@ def data_description_file(outdir):
                     'generating it with provided info')
         utils.writejson(file_path, data_dict)
 
+
+def README_file(outdir):
+    """
+    Create README file if it does not exist.
+    If it exists, do nothing.
+
+    Parameters
+    ----------
+    outdir: path
+        Full path to the output directory.
+
+    """
+    file_path = os.path.join(outdir, 'README')
+    text = 'Empty README, plese fill in describing the dataset in more details SHOULD be provided'
+    if not os.path.exists(file_path):
+        LGR.warning('phys2bids could not find README,'
+                    'generating it EMPTY, please fill in the necessary info')
+        utils.writefile(file_path, '', text)
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -242,3 +242,18 @@ def participants_file(outdir, yml, sub):
             participants_data = ['n/a'] * header_length
             participants_data[p_id_idx] = f'sub-{sub}'
             utils.append_list_as_row(file_path, participants_data)
+
+
+def data_description_file(outdir, name):
+    """
+    Create dataset_description.json file if it does not exist.
+    If it exists, do nothing.
+
+    Parameters
+    ----------
+    outdir: path
+        Full path to the output directory.
+    name: str
+        Dataset name
+
+    """
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -277,7 +277,7 @@ def readme_file(outdir):
         Full path to the output directory.
 
     """
-    file_path = os.path.join(outdir, 'README')
+    file_path = os.path.join(outdir, 'README.md')
     text = 'Empty README, please fill in describing the dataset in more detail.'
     if not os.path.exists(file_path):
         LGR.warning('phys2bids could not find README,'
--- a/phys2bids/bids.py
+++ b/phys2bids/bids.py
@@ -266,7 +266,7 @@ def dataset_description_file(outdir):
         utils.writejson(file_path, data_dict)
 
 
-def README_file(outdir):
+def readme_file(outdir):
     """
     Create README file if it does not exist.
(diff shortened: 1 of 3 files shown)


vinferrer:
@eurunuela, @smoia I think this is ready for review

phys2bids/bids.py
```diff
@@ -242,3 +242,44 @@ def participants_file(outdir, yml, sub):
...
+                 "BIDSVersion": "1.4.0", "DatasetType": "raw"}
+    file_path = os.path.join(outdir, 'dataset_description.json')
+    # check if dataset_description.json exists, if it doesn't create it
+    if not os.path.exists(file_path):
+        LGR.warning('phys2bids could not find dataset_description.json,'
+                    'generating it with provided info')
+        utils.writejson(file_path, data_dict)
+
+
+def README_file(outdir):
```
eurunuela, line 269:
```suggestion
def readme_file(outdir):
```

phys2bids/phys2bids.py
```diff
@@ -269,6 +270,10 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
         # Update the file if the subject is not in the file.
         # Do not update if the subject is already in the file.
         participants_file(outdir, yml, sub)
+        # Generate dataset_description.json file if it doesn't exist already.
+        dataset_description_file(outdir)
+        # Generate README file if it doesn't exist already.
+        README_file(outdir)
```
eurunuela, line 276:
```suggestion
        readme_file(outdir)
```

phys2bids/phys2bids.py
```diff
@@ -35,6 +35,7 @@
 
 from phys2bids import utils, viz, _version
 from phys2bids.bids import bidsify_units, use_heuristic, participants_file
+from phys2bids.bids import dataset_description_file, README_file
```
eurunuela, line 38:
```suggestion
from phys2bids.bids import dataset_description_file, readme_file
```

phys2bids/tests/test_bids.py
```diff
@@ -4,7 +4,7 @@
 
 import pytest
 
-from phys2bids.bids import bidsify_units, use_heuristic
+from phys2bids.bids import bidsify_units, use_heuristic, README_file, dataset_description_file
```
eurunuela, line 7:
```suggestion
from phys2bids.bids import bidsify_units, use_heuristic, readme_file, dataset_description_file
```

phys2bids/tests/test_bids.py
```diff
@@ -58,3 +58,17 @@ def test_use_heuristic(tmpdir, test_sub, test_ses):
                    f'_task-test_rec-biopac_run-01_recording-test_physio')
 
     assert os.path.normpath(test_result) == os.path.normpath(str(heur_path))
+
+
+@pytest.mark.parametrize('outdir', '.')
+def test_README_file(outdir):
+    README_file(outdir)
```
eurunuela, line 65:
```suggestion
    readme_file(outdir)
```

phys2bids/bids.py
```diff
@@ -242,3 +242,44 @@ def participants_file(outdir, yml, sub):
...
+    If it exists, do nothing.
+
+    Parameters
+    ----------
+    outdir: path
+        Full path to the output directory.
+
+    """
+    file_path = os.path.join(outdir, 'README')
+    text = 'Empty README, plese fill in describing the dataset in more details SHOULD be provided'
```
eurunuela, line 281:
```suggestion
    text = 'Empty README, please fill in describing the dataset in more detail.'
```

phys2bids/phys2bids.py
smoia: At this point, we're importing a lot from bids. It would be better to add "bids" to the import of phys2bids (two lines above) and update the code consequently!

phys2bids/bids.py
smoia: Nothing special, but I'd move this in the if statement

phys2bids/bids.py
```diff
@@ -242,3 +242,44 @@ def participants_file(outdir, yml, sub):
...
+    If it exists, do nothing.
+
+    Parameters
+    ----------
+    outdir: path
+        Full path to the output directory.
+
+    """
+    # dictionary that will be written for the basic dataset description version
+    data_dict = {"Name": os.path.splitext(os.path.basename(outdir))[0],
```
smoia, line 259: I would check first if the file exists. If it does, we could read it and update the missing fields and the bids version. If it does not, we could create it.

vinferrer:
Changes implemented guys

phys2bids/bids.py
vinferrer: I wouldn't do that, All the fields there are Required except datatype, therefore there is no need to update for the moment. A different thing is that we add new fields in the future

smoia: Fair enough. Can you move this into the if statement then?

phys2bids/phys2bids.py
vinferrer: done

phys2bids/phys2bids.py
```diff
@@ -35,6 +35,7 @@
 
 from phys2bids import utils, viz, _version
 from phys2bids.bids import bidsify_units, use_heuristic, participants_file
+from phys2bids.bids import dataset_description_file, readme_file
```
smoia: At this point, it would be better to:
```suggestion
from phys2bids import utils, viz, _version, bids
```
And then adapt the rest of the code

phys2bids/bids.py
vinferrer: Plus we cannot update the bids version of the whole dataset because we are not ensuring bids specs from other kind of data

phys2bids/bids.py
```diff
@@ -242,3 +242,44 @@ def participants_file(outdir, yml, sub):
...
+                 "BIDSVersion": "1.4.0", "DatasetType": "raw"}
+    file_path = os.path.join(outdir, 'dataset_description.json')
+    # check if dataset_description.json exists, if it doesn't create it
+    if not os.path.exists(file_path):
+        LGR.warning('phys2bids could not find dataset_description.json,'
```
smoia, line 264:
```suggestion
    file_path = os.path.join(outdir, 'dataset_description.json')
    # check if dataset_description.json exists, if it doesn't create it
    if not os.path.exists(file_path):
        data_dict = {"Name": os.path.splitext(os.path.basename(outdir))[0],
                 "BIDSVersion": "1.4.0", "DatasetType": "raw"}
        LGR.warning('phys2bids could not find dataset_description.json,'
```

PR #259, opened 2020-06-19
Remove redundant integration tests

### LINKED ISSUE #258: Reduce the amount of integration tests to make tests faster
I'm opening this discussion because I think that we could probably get rid of a couple of the integration tests for the sake of speed in our testing. Integration tests are the ones that take the longer because they run the entire pipeline and there's a lot of file reading and checking, which is always slow.

So, I suggest we remove the [ACQ](https://github.com/physiopy/phys2bids/blob/8d42c28107574b858ed4033dd8f1c79b690753be/phys2bids/tests/test_integration.py#L112) and [TXT](https://github.com/physiopy/phys2bids/blob/8d42c28107574b858ed4033dd8f1c79b690753be/phys2bids/tests/test_integration.py#L59) integration tests. We already have an integration test for multi-frequency files, which goes through the same pipeline as the other and the file is much lighter, hence faster.

I will open a draft PR to see if we would have a big decrease in coverage (which I don't expect), but I'd like to know what you think about it.

eurunuela:
Closes #258

## Proposed Changes

  - Remove redundant integration tests to gain in testing speed.

### DIFF
--- a/phys2bids/heuristics/heur_test_acq.py
+++ b/phys2bids/heuristics/heur_test_acq.py
@@ -56,7 +56,7 @@ def heur(physinfo):
     # ##  See example below          ## #
     # ################################# #
 
-    if fnmatch.fnmatchcase(physinfo, '*samefreq*'):
+    if fnmatch.fnmatchcase(physinfo, '*onescan*'):
         info['task'] = 'test'
         info['run'] = '01'
         info['rec'] = 'biopac'
--- a/phys2bids/heuristics/heur_test_multifreq.py
+++ b/phys2bids/heuristics/heur_test_multifreq.py
@@ -56,7 +56,7 @@ def heur(physinfo):
     # ##  See example below          ## #
     # ################################# #
 
-    if fnmatch.fnmatchcase(physinfo, '*samefreq*'):
+    if fnmatch.fnmatchcase(physinfo, '*onescan*'):
         info['task'] = 'test'
         info['run'] = '01'
         info['rec'] = 'biopac'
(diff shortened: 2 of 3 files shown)


eurunuela:
I don't know what's going on with Travis... some numpy related issues appeared and I honestly don't know why. Nothing has changed in the installation process... any idea @rmarkello ?

Anyway, the tests are much faster when removing the redundant ones (see [here](https://travis-ci.org/github/physiopy/phys2bids/jobs/699984185)). I checked codecov and we would have a -0.52% decrease in coverage (to 92.97%). The interfaces folder would still be covered at 98.83%, which means that removing the ACQ and TXT integration tests has no impact on `txt.py` and `acq.py`.

This means that the only ones below 90% would be `bids.py` (which will probably be above 90% when #254 is merged), `due.py` and `phys2bids.py`. I think it should be easy to increase coverage on them with silly unit tests without compromising testing speed.

So, I think merging this PR would improve our testing suite.

vinferrer:
LGTM, i think this is ready for merge

eurunuela:
We will have a bit of a decrease in coverage (down to 91%) but I expect #254 to increase that (bids.py from 73% to 98%).

eurunuela:
**Just to keep this written somewhere:**

As per discussion over Zoom with @smoia and @vinferrer , I added the acq test back and merged the others into one single test (heuristic with multi-freq txt file and checks on the logger).  The heuristic one takes 8 seconds on my laptop while the acq one takes around 40 seconds. I'll check if we have a short acq file on OSF and update if I find one. 

Anyway, this is a very big decrease in testing time.

PR #260, opened 2020-06-19
Change os path import and linter

smoia:
Closes nothing.
It's just a little lintering (I know Travis isn't working today, but let's still check the linter test, since we don't have it in Azure), while removing unused imports.
In addition, instead of importing `os` or its modules twice, it's importing `remove` and the three functions that we use from `path`.
I know this is not standard `os` import (@rmarkello you know better than us), but I think it's a bit better rather than importing os twice using similar modules, if we want a solution that saves digits.
I'd prefer to stick with `os.path.function` since it's easier to read, but if we want shorter solutions, at least it's a tiny bit clearer.

## Proposed Changes

  - Change `os` and `os.path` import
  - Remove csv import
  - General linter

### DIFF
--- a/phys2bids/tests/test_integration.py
+++ b/phys2bids/tests/test_integration.py
@@ -1,12 +1,11 @@
 import glob
 import json
 import math
-import os
 import re
 import shutil
 import subprocess
-from csv import reader
-from os.path import join as opj
+from os import remove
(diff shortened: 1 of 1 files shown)


eurunuela:
Thanks @smoia !

PR #275, opened 2020-07-07
Correct trigger channel indexing while reading AcqKnowledge files.

### LINKED ISSUE #272: Trigger channel index is not incremented in acq populate_phys_input
# Updated Issue

The acq-generated `BlueprintInput` objects' `trigger_index` attribute does not correspond to the actual trigger channel index.
The trigger index needs to be incremented down in the acq version of `populate_phys_input()`.

# Original Issue

When time is added as a channel to the time series data, the trigger channel index needs to be incremented, or else it will not reference the trigger channel. This is done in the [txt version of populate_phys_input](https://github.com/physiopy/phys2bids/blob/a2493bf383d3a7668c549099215bd3ddd60f8831/phys2bids/interfaces/txt.py#L152), but not in the the [acq version](https://github.com/physiopy/phys2bids/blob/a2493bf383d3a7668c549099215bd3ddd60f8831/phys2bids/interfaces/acq.py#L51).

## Expected Behavior

The `BlueprintInput` object created from reading in a `.acq` file should have a `.trigger_idx` attribute pointing to the trigger channel.

## Actual Behavior

`BlueprintInput` objects initialized from `.acq` files have `.trigger_idx` attributes pointing to the channel right _before_ the trigger channel.

## Steps to Reproduce the Problem

1. Read an .acq file.
2. Compare the `chtrig` argument from the `populate_phys_input()` call to the `.trigger_idx` attribute of the generated `BlueprintInput` object. It should have a value of `chtrig + 1`, not `chtrig`.

## Specifications

    - Python version: 3.7
    - phys2bids version: current master at a2493bf
    - Platform: OSX

## Possible solution

Add 1 to chtrig when initializing the BlueprintInput object in the acq version of populate_phys_input.

smoia:
Closes #272, although the fixed problem is not the same - still, thanks @tsalo!

Turns out the algorithm was meant to extract the channel _after_ the one of the trigger since we modified phys2bids to have `-chtrig` human-friendly (rather than pythonic). 

## Proposed Changes

  - the `chtrig` index is lowered by one in place when extracting the timeseries from the trigger channel.
  - the docstring now indicate that chtrig is meant to represent an index starting from 1.

### DIFF
--- a/phys2bids/interfaces/acq.py
+++ b/phys2bids/interfaces/acq.py
@@ -30,9 +30,9 @@ def populate_phys_input(filename, chtrig):
 
     Note
     ----
-    chtrig is not a python index - instead, it's human readable.
-    This is handy because when initialising the class, it gets initiliased
-    with a channel more at the beginning - that is already taken into account!
+    chtrig is not a 0-based Python index - instead, it's human readable (i.e., 1-based).
+    This is handy because, when initialising the class, a new channel corresponding
+    to time is added at the beginning - that is already taken into account!
 
     See Also
     --------
--- a/phys2bids/interfaces/acq.py
+++ b/phys2bids/interfaces/acq.py
@@ -21,23 +21,29 @@ def populate_phys_input(filename, chtrig):
     filename: str
         path to the txt labchart file
     chtrig : int
-        index of trigger channel
+        index of trigger channel.
+        !!! ATTENTION: IT'S MEANT TO REPRESENT AN INDEX STARTING FROM 1 !!!
 
     Returns
     -------
     BlueprintInput
 
+    Note
+    ----
+    chtrig is not a python index - instead, it's human readable.
+    This is handy because when initialising the class, it gets initiliased
+    with a channel more at the beginning - that is already taken into account!
+
     See Also
     --------
     physio_obj.BlueprintInput
     """
-
     with warnings.catch_warnings():
         warnings.filterwarnings('ignore', category=DeprecationWarning)
         data = read_file(filename).channels
 
-    freq = [data[chtrig].samples_per_second, ]
-    timeseries = [data[chtrig].time_index, ]
+    freq = [data[chtrig - 1].samples_per_second, ]
+    timeseries = [data[chtrig - 1].time_index, ]
     units = ['s', ]
     names = ['time', ]
 
--- a/phys2bids/interfaces/acq.py
+++ b/phys2bids/interfaces/acq.py
@@ -21,7 +21,7 @@ def populate_phys_input(filename, chtrig):
     filename: str
         path to the txt labchart file
     chtrig : int
-        index of trigger channel.
+        index of trigger channel. Must be a positive, non-zero integer.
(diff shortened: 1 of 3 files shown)


tsalo:
The CLI's docstring and default value for `chtrig` also indicate zero is the first channel. I would update those as well as add a check for either the parser's `chtrig` parameter or at the beginning of the `phys2bids` Python function to ensure that the `chtrig` value is an integer greater than or equal to 1.

phys2bids/interfaces/acq.py
```diff
@@ -21,23 +21,29 @@ def populate_phys_input(filename, chtrig):
...
 
     Returns
     -------
     BlueprintInput
 
+    Note
+    ----
+    chtrig is not a python index - instead, it's human readable.
+    This is handy because when initialising the class, it gets initiliased
+    with a channel more at the beginning - that is already taken into account!
```
tsalo, line 35:
```suggestion
    chtrig is not a 0-based Python index - instead, it's human readable (i.e., 1-based).
    This is handy because, when initialising the class, a new channel corresponding
    to time is added at the beginning - that is already taken into account!
```

phys2bids/interfaces/acq.py
```diff
@@ -21,23 +21,29 @@ def populate_phys_input(filename, chtrig):
     filename: str
         path to the txt labchart file
     chtrig : int
-        index of trigger channel
+        index of trigger channel.
```
tsalo, line 24:
```suggestion
        index of trigger channel. Must be an integer >= 1.
```

phys2bids/interfaces/acq.py
tsalo: You don't need to accept this suggestion. I think it would be sufficient to just clarify that the new channel reflects time. Also, `initialised` was misspelled as `initiliased` (which might have gotten buried in the suggestion).

phys2bids/phys2bids.py
```diff
@@ -191,6 +191,9 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     # Check options to make them internally coherent pt. II
     # #!# This can probably be done while parsing?
     indir = utils.check_input_dir(indir)
+    if chtrig < 1:
+        raise Exception('Wrong trigger cahnnel. Channel indexing starts with 1!')
```
tsalo, line 195:
```suggestion
        raise Exception('Wrong trigger channel. Channel indexing starts with 1!')
```
Just a minor typo.

vinferrer:
LGTM!!! althought remmeber bugs PR only need one reviewer

tsalo:
@vinferrer I didn't realize that the `physiopy` community classified PRs like that, but that's awesome! In [the docs](https://phys2bids.readthedocs.io/en/latest/contributorfile.html#pr), though, it looks like two reviews are currently required for bug fixes. Perhaps that needs to be updated?

> 2. Have the necessary amount of approving reviews, even if you’re a long time contributor. You can ask one (or more) contributor to do that review, if you think they align more with the content of your PR. You need **one** review for documentation, tests, and small changes, and **two** reviews for bugs, refactoring and enhancements.

PR #286, opened 2020-08-13
Add CircleCI for automatic testing and correct integration test parameters

eurunuela:
Closes #

## Proposed Changes

  - Set up CircleCI to be the platform where we run automatic tests.
  - Correct some input parameters for integration tests.
  - Mark trigger plot test as "expected to fail" until we fix the savefig function.
  - Only save figure if not running tests (we should undo that when we fix the issue).


vinferrer:
LGTM!!

tsalo:
I'd recommend adding a docs build job to the workflow. You'll need the [`circle-artifacts-redirector` GitHub App](https://github.com/apps/circleci-artifacts-redirector/).

dev_requirements.txt
```diff
@@ -0,0 +1,13 @@
+-r requirements.txt
```
tsalo, line 1: I would recommend dropping this extra file and just using the extra_requires options in `setup.cfg`. Essentially, just install within the config with `pip install -e .[tests]`. You can use the same for RTD (see [NiMARE's example](https://github.com/neurostuff/NiMARE/blob/master/.readthedocs.yml)).

smoia:
> I'd recommend adding a docs build job to the workflow. You'll need the [`circle-artifacts-redirector` GitHub App](https://github.com/apps/circleci-artifacts-redirector/).

It's installed now!

dev_requirements.txt
tsalo: That would also mean dropping `requirements.txt` and changing your checksums from `requirements.txt` + `dev_requirements.txt` to `setup.cfg`.

eurunuela:
@tsalo I had that in mind but wanted to implement the unit and integration tests before. Hopefully CircleCI will be more stable than Travis.

dev_requirements.txt
eurunuela: Thans @tsalo ! Trying to make that work now. Help appreciated!

eurunuela:
I can't seem to find the way to make the docs and style tests work. Any idea @tsalo @smoia ?

eurunuela:
I managed to make the docs work. Apparently, it really needed the environment to finish before starting the job. No idea why the lint fails though...

tsalo:
Great work!

Two small things I noticed:
- You can remove `docs/requirements.txt` now that there's a readthedocs YAML file pointing to `setup.cfg`.
- You'll need to add an `.circleci/artifact_path` file like [this one](https://github.com/neurostuff/NiMARE/blob/master/.circleci/artifact_path) so that the artifact redirector can create a status check with a link to the rendered documentation.

eurunuela:
Thank you so much @tsalo ! That was very helpful.

> EDIT: As for the style check, I think all of the D[XXX] errors you're getting are from the doctests. You can disable that by removing `doctest = True` from the flake8 config in `setup.cfg`. Otherwise, we're looking at fixing all of the docstring tests.

Looks like we're gonna have to open a PR to fix those.

PR #283, opened 2020-08-11
Start of sampling time is now the same for all frequencies

vinferrer:
Closes #282 

## Proposed Changes

  - Now the sampling start time is the same for all frequencies blueprint output objects.
  - time channel is now in all frequencies blueprint  output objects down-sampled to the appropriate frequency

PR #288, opened 2020-08-18
Add (resampled) trigger to all output files 

### LINKED ISSUE #274: Add (resampled) trigger channel in every multifreq output
This comes from #219. It would be a good idea (not mandatory, but good) to add a refactored trigger channel in each output in case of multiple outputs due to multifrequency files.
At the moment we're not doing it!

## Context / Motivation

It would help improving (visual) comparison between traces and fMRI trigger!

## Possible Implementation

- [ ] Create a resampled trigger channel for each (unique) frequency encountered in `BlueprintInput.freq` list (e.g. with `set(BlueprintInput.freq)` or `np.unique(BlueprintInput.freq)`) and append it to the channel list. This could be a class method.

vinferrer:
Closes #274 

## Proposed Changes

  - Now the resampled trigger is in all the frequency files outputs when a multifrequency file is inputed


phys2bids/tests/test_integration.py
```diff
@@ -45,7 +45,7 @@ def test_integration_acq(skip_integration, samefreq_full_acq_file):
         assert isfile(join(test_path, 'Test_belt_pulse_samefreq' + suffix))
 
     # Check files in extra are generated
-    for suffix in ['.log']:  #, '_trigger_time.png']:
+    for suffix in ['.log']:  # , '_trigger_time.png']:
```
eurunuela, line 48:
```suggestion
    for suffix in ['.log']:
```

phys2bids/phys2bids.py
```diff
@@ -373,6 +373,22 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
             # Also create a BlueprintOutput object for each unique frequency found.
             # Populate it with the corresponding blueprint input and replace it
             # in the dictionary.
+            # add time channel in the proper frequency
+            if uniq_freq != phys_in[run].freq[0]:
+                phys_out[key].ch_name.insert(0, phys_in[run].ch_name[0])
+                phys_out[key].units.insert(0, phys_in[run].units[0])
+                phys_out[key].timeseries.insert(0, np.linspace(phys_in[run].timeseries[0][0],
+                                                phys_in[run].timeseries[0][-1],
+                                                num=phys_out[key].timeseries[0].shape[0]))
```
eurunuela, line 382: This will be merged in #283 . Is it really necessary to have both PRs? If so, we're going to have to wait for that one to be merged.

phys2bids/phys2bids.py
```diff
@@ -373,6 +373,22 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
...
+            # add time channel in the proper frequency
+            if uniq_freq != phys_in[run].freq[chtrig]:
+                phys_out[key].ch_name.insert(1, phys_in[run].ch_name[chtrig])
+                phys_out[key].units.insert(1, phys_in[run].units[chtrig])
+                new_time = np.linspace(phys_in[run].timeseries[0][0],
```
eurunuela, line 387: I'd call this variable something like `resampled_time` or something like that, to make it more descriptive.

phys2bids/phys2bids.py
```diff
@@ -373,6 +373,22 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
...
             # Populate it with the corresponding blueprint input and replace it
             # in the dictionary.
+            # add time channel in the proper frequency
+            if uniq_freq != phys_in[run].freq[0]:
+                phys_out[key].ch_name.insert(0, phys_in[run].ch_name[0])
+                phys_out[key].units.insert(0, phys_in[run].units[0])
+                phys_out[key].timeseries.insert(0, np.linspace(phys_in[run].timeseries[0][0],
+                                                phys_in[run].timeseries[0][-1],
+                                                num=phys_out[key].timeseries[0].shape[0]))
+            # add time channel in the proper frequency
```
eurunuela, line 383: I think the comment should be more explicit. It is an exact copy of the one above.

phys2bids/phys2bids.py
vinferrer: both PRs have been merge already in my forks that's why is duplicated, but yeah #283 will be merged first

vinferrer:
Changes made, @smoia you are the assigned reviewer since this is your code part

phys2bids/phys2bids.py
```diff
@@ -380,6 +380,16 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
                 phys_out[key].timeseries.insert(0, np.linspace(phys_in[run].timeseries[0][0],
                                                 phys_in[run].timeseries[0][-1],
                                                 num=phys_out[key].timeseries[0].shape[0]))
+            # add trigger channel in the proper frequency
```
eurunuela, line 383:
```suggestion
            # Add trigger channel in the proper frequency.
```

eurunuela (changes requested):

phys2bids/phys2bids.py
vinferrer: I think this changed when merging with master, sry about it

vinferrer:
Sry about that, it changed when mergin with master

eurunuela (approved):
LGTM!

vinferrer:
I have notice that the phys2bids unit-test isn't triggering any of the if statements, leaving all the coverage to the integration test. Should we add them to ``test_phys2bids.py``?

eurunuela:
Yes. We should add some checks to those if statements in `test_phys2bids.py. Will you open an issue for it?

phys2bids/phys2bids.py
```diff
@@ -377,6 +377,16 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
                 phys_out[key].timeseries.insert(0, np.linspace(phys_in[run].timeseries[0][0],
                                                 phys_in[run].timeseries[0][-1],
                                                 num=phys_out[key].timeseries[0].shape[0]))
+            # Add trigger channel in the proper frequency.
+            if uniq_freq != phys_in[run].freq[chtrig]:
+                phys_out[key].ch_name.insert(1, phys_in[run].ch_name[chtrig])
+                phys_out[key].units.insert(1, phys_in[run].units[chtrig])
+                resampled_time = np.linspace(phys_in[run].timeseries[0][0],
```
smoia, line 384: I'm probably missing something, but isn't `resampled_time` exactly the same as `phys_out[key].timeseries[0]` at this point in the code?

If the frequency the loop is in is the same as the one of the "time" channel, then `phys_out[key].timeseries[0] == phys_in[key].timeseries[0]`, and this linspace creates a perfect copy of `phys_out[key].timeseries[0]`.
If the frequency the loop is in is not the same, then lines 384-386 are the same as the second argument in `insert` in lines 377-379.

If so, wouldn't it be better to use `phys_out[key].timeseries[0]` straight away rather than recomputing it not more than five lines later?

I guess that then these three lines can be skipped, and line 387 can become `phys_out[key].timeseries.insert(1, np.interp(phys_out[key].timeseries[0], `

smoia (changes requested):
It looks good, but can you address my only comment?
I'd like to know if I'm missing something and what I'm proposing is not better than the current solution.

phys2bids/phys2bids.py
vinferrer: You are completely correct, changes on the way

smoia (approved):
LGTM!

PR #291, opened 2020-08-19
Add tests for exceptions

### LINKED ISSUE #276: Testing error raises
We should add testing of those cases that we know should break the code - basically, to check that we raise exceptions when we should.

## Context / Motivation

Why? Because it's fun! And also because it reduces even more the possibility of mistakes.

## Possible Implementation

- [ ] (necessary) Cover all the `raise Exception` statements with [pytest](https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions)
- [ ] (optional) Think about cases that might break the code and check that they do. If they don't, open `bug` issues.

eurunuela:
Closes #276

## Proposed Changes

  - Add tests for all the raise statements along the phys2bids package.


phys2bids/tests/test_utils.py
```diff
@@ -30,6 +32,10 @@ def test_check_input_ext():
 def test_check_input_type(testpath, samefreq_full_acq_file):
     assert utils.check_input_type(samefreq_full_acq_file, testpath)
 
+    with raises(Exception) as errorinfo:
+        utils.check_input_type('nobel_prize.win', testpath)
+    assert 'wasn\'t found' in str(errorinfo.value)
```
tsalo, line 37:
```suggestion
    assert "wasn't found" in str(errorinfo.value)
```

phys2bids/tests/test_phys2bids.py
```diff
@@ -41,3 +42,23 @@ def test_print_json(tmpdir):
...
+    with raises(Exception) as errorinfo:
+        phys2bids.phys2bids(filename=test_filename,chtrig=0)
+    assert 'Wrong trigger' in str(errorinfo.value)
+
+    test_path, test_filename = os.path.split(samefreq_full_acq_file)
+    with raises(Exception) as errorinfo:
+        phys2bids.phys2bids(filename=test_filename, num_timepoints_expected=[70], tr=[1.3, 2],
+                            indir=test_path, outdir=test_path)
+    assert 'doesn\'t match' in str(errorinfo.value)
+
```
vinferrer, line 59: what is the propouse of this?

tsalo: Could you clarify what you mean? Are you talking about the backslash?

phys2bids/tests/test_phys2bids.py
```diff
@@ -41,3 +42,23 @@ def test_print_json(tmpdir):
...
+    with raises(Exception) as errorinfo:
+        phys2bids.phys2bids(filename=test_filename,chtrig=0)
+    assert 'Wrong trigger' in str(errorinfo.value)
+
+    test_path, test_filename = os.path.split(samefreq_full_acq_file)
+    with raises(Exception) as errorinfo:
+        phys2bids.phys2bids(filename=test_filename, num_timepoints_expected=[70], tr=[1.3, 2],
+                            indir=test_path, outdir=test_path)
+    assert 'doesn\'t match' in str(errorinfo.value)
```
tsalo, line 58:
```suggestion
    assert "doesn't match" in str(errorinfo.value)
```

phys2bids/tests/test_phys2bids.py
vinferrer: No, no, I wasn't properly understanding how errorinfo.value worked. My apologies

vinferrer (approved):
LGTM!!!

phys2bids/phys2bids.py
```diff
@@ -222,6 +222,7 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     elif ftype == 'txt':
         from phys2bids.interfaces.txt import populate_phys_input
     else:
+        # This is never called.
```
tsalo, line 225: Any particular reason for this? Is the issue that it's not reachable or that, in practice, no one will actually try running a file without `.txt.` or `.acq`?

phys2bids/tests/test_phys2bids.py
```diff
@@ -41,3 +42,22 @@ def test_print_json(tmpdir):
         loaded_data = json.load(src)
 
     assert test_json_data == loaded_data
+
+
+def test_raise_exception(samefreq_full_acq_file):
+    test_filename = 'input.txt'
+
+    with raises(Exception) as errorinfo:
```
tsalo, line 50: This isn't a concern for this PR, but in general it would be better to have more specific exception types. There would be less reason to check the exception message as well (since checking the exception type would be sufficiently specific).

phys2bids/tests/test_phys2bids.py
```diff
@@ -41,3 +42,22 @@ def test_print_json(tmpdir):
...
+        phys2bids.phys2bids(filename=test_filename,chtrig=0)
```
tsalo, line 51:
```suggestion
        phys2bids.phys2bids(filename=test_filename, chtrig=0)
```

phys2bids/phys2bids.py
eurunuela: It is not reachable as [this](https://github.com/physiopy/phys2bids/blob/c20cef844f3b22741635a8d0f2178433c6a548c8/phys2bids/utils.py#L107) exception will be raised before ([see here](https://github.com/physiopy/phys2bids/blob/c20cef844f3b22741635a8d0f2178433c6a548c8/phys2bids/phys2bids.py#L197)) getting to this point.

phys2bids/tests/test_phys2bids.py
eurunuela: I completely agree. I'll open an issue to work on this.

phys2bids/phys2bids.py
eurunuela: I didn't remove that part in this PR because it's meant to tackle the tests only. We could do it in the MATLAB interface PR if we agree that's the way to go (I think it should be done).

tsalo: Thanks. Yes, it should totally be removed. Plus that's two untested lines being dropped (yay coverage increase!).

phys2bids/tests/test_phys2bids.py
tsalo: Thanks!

phys2bids/phys2bids.py
eurunuela: My latest commit should take care of this. Thanks for the review!

PR #292, opened 2020-08-19
Add support for MATLAB files

### LINKED ISSUE #25: Add support for matlab files
## Detailed Description

We have some labchart-exported [matlab files in our OSF](https://osf.io/2j43t/). We could implement an interface function that reads matlab files, in order to support yet another type of data!

## Possible Implementation

- Look at the functions in [`interfaces`](https://github.com/physiopy/phys2bids/tree/master/phys2bids/interfaces)
- Using [scipy](https://docs.scipy.org/doc/scipy/reference/tutorial/io.html) or [h5py](https://www.h5py.org/) (or both), write an itnerface function to read Labchart data in matlab format. 
- Write a test for such interface function!

eurunuela:
Closes #25

## Proposed Changes

  - Create a MATLAB interface to read MATLAB files and transform them into `BlueprintInput` objects.
  - Add unit tests for the MATLAB interface.
  - Add integration test for a MATLAB file.

### DIFF
--- a/phys2bids/interfaces/mat.py
+++ b/phys2bids/interfaces/mat.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""phys2bids interface for MATLAB files."""
+
+import logging
+
+import numpy as np
+from scipy.io import loadmat
+
+from phys2bids.physio_obj import BlueprintInput
+
+LGR = logging.getLogger(__name__)
+
+
+def populate_phys_input(filename, chtrig):
+    """
+    Populate object phys_input from MATLAB files.
+
+    Parameters
+    ----------
+    filename: str
+        path to the txt labchart file
+    chtrig : int
+        index of trigger channel.
+        !!! ATTENTION: IT'S MEANT TO REPRESENT AN INDEX STARTING FROM 1 !!!
+
+    Returns
+    -------
+    BlueprintInput
+
+    Note
+    ----
+    chtrig is not a 0-based Python index - instead, it's human readable (i.e., 1-based).
+    This is handy because, when initialising the class, a new channel corresponding
+    to time is added at the beginning - that is already taken into account!
+
+    See Also
+    --------
+    physio_obj.BlueprintInput
+    """
+
+    # Load MATLAB file into dictionary.
+    mat_dict = loadmat(filename)
+
+    # Convert data into 1d numpy array for easier indexing.
+    data = np.squeeze(np.asarray(mat_dict['data']))
+
+    # Extract number of channels.
+    n_channels = len(mat_dict['titles'])
+
+    # Stores MATLAB data into lists.
+    timeseries = [data[:int(mat_dict['dataend'][0])], ]
+    freq = [mat_dict['samplerate'][ch][0] for ch in range(n_channels)]
+    units = [mat_dict['unittext'][int(mat_dict['unittextmap'][ch][0] - 1)].strip() for ch in range(n_channels)]
+    names = [mat_dict['titles'][ch].strip() for ch in range(n_channels)]
+
+    # Appends data from channels > 1.
+    if n_channels > 1:
+        for ch in range(n_channels - 1):
+            idx_start = int(mat_dict['dataend'][ch]) + 1
+            idx_end = int(mat_dict['dataend'][ch+1])
+            timeseries.append(data[idx_start:idx_end])
+
+    return BlueprintInput(timeseries, freq, names, units, chtrig)
--- a/phys2bids/interfaces/mat.py
+++ b/phys2bids/interfaces/mat.py
@@ -46,20 +46,28 @@ def populate_phys_input(filename, chtrig):
     # Convert data into 1d numpy array for easier indexing.
     data = np.squeeze(np.asarray(mat_dict['data']))
 
-    # Extract number of channels.
+    # Extract number of channels and tick rate.
     n_channels = len(mat_dict['titles'])
+    t_freq = mat_dict['tickrate'][0][0]
 
     # Stores MATLAB data into lists.
-    timeseries = [data[:int(mat_dict['dataend'][0])], ]
-    freq = [mat_dict['samplerate'][ch][0] for ch in range(n_channels)]
-    units = [mat_dict['unittext'][int(mat_dict['unittextmap'][ch][0] - 1)].strip() for ch in range(n_channels)]
-    names = [mat_dict['titles'][ch].strip() for ch in range(n_channels)]
+    timeseries = []
+    freq = [t_freq, ]
+    units = ['s', ]
+    names = ['time', ]
 
-    # Appends data from channels > 1.
-    if n_channels > 1:
-        for ch in range(n_channels - 1):
-            idx_start = int(mat_dict['dataend'][ch]) + 1
-            idx_end = int(mat_dict['dataend'][ch+1])
-            timeseries.append(data[idx_start:idx_end])
+    for ch in range(n_channels):
+        units.append(mat_dict['unittext'][int(mat_dict['unittextmap'][ch][0] - 1)].strip())
+        names.append(mat_dict['titles'][ch].strip())
+        freq.append(mat_dict['samplerate'][ch][0])
+        idx_start = int(mat_dict['datastart'][ch])
+        idx_end = int(mat_dict['dataend'][ch])
+        timeseries.append(data[idx_start:idx_end])
+
+    # Calculate duration based on frequency and create time channel.
+    interval = 1 / t_freq
+    duration = (timeseries[0].shape[0] + 1) * interval
+    t_ch = np.ogrid[0:duration:interval][:-1]
+    timeseries = [t_ch, ] + timeseries
 
     return BlueprintInput(timeseries, freq, names, units, chtrig)
--- a/phys2bids/interfaces/mat.py
+++ b/phys2bids/interfaces/mat.py
@@ -39,14 +39,13 @@ def populate_phys_input(filename, chtrig):
     --------
     physio_obj.BlueprintInput
     """
-
     # Load MATLAB file into dictionary.
     mat_dict = loadmat(filename)
 
     # Convert data into 1d numpy array for easier indexing.
     data = np.squeeze(np.asarray(mat_dict['data']))
 
-    # Extract number of channels and tick rate.
+    # Extract number of channels and tick rate.
     n_channels = len(mat_dict['titles'])
     t_freq = mat_dict['tickrate'][0][0]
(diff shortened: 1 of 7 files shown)


eurunuela:
@RayStick it would be great if you could make sure that the inputs to phys2bids in the integration test are correct; i.e. `tr=1.2` and `num_timepoints_expected=100`. 

I think `num_timepoints_expected` is correct but I have no idea about `tr`.

tsalo:
Is this a good chance to talk about the organization of these I/O functions? I find the `phys2bids.interfaces.[modality].populate_phys_input()` structure unnecessarily confusing. Why not have `phys2bids.io.load_matlab()`, `phys2bids.io.load_text()`, and `phys2bids.io.load_acq()`?

eurunuela:
> Is this a good chance to talk about the organization of these I/O functions? I find the `phys2bids.interfaces.[modality].populate_phys_input()` structure unnecessarily confusing. Why not have `phys2bids.io.load_matlab()`, `phys2bids.io.load_text()`, and `phys2bids.io.load_acq()`?

I completely agree and did think of it while working this PR. I think it's a very good point we should bring to our next developer call. However, I don't think that should stop this PR from being merged. Actually, it will probably be easier to work on those changes once this PR is merged.

tsalo:
Thanks @eurunuela. You're right- this function can always be moved/renamed at a later date. There's no need to hold up the PR based on that. I'll just open a separate issue about it.

vinferrer:
> 
> 
> @RayStick it would be great if you could make sure that the inputs to phys2bids in the integration test are correct; i.e. `tr=1.2` and `num_timepoints_expected=100`.
> 
> I think `num_timepoints_expected` is correct but I have no idea about `tr`.

You know, I think you can check that in OSF at least in some of the files

vinferrer:
In any case, seems solid code.

eurunuela:
> You know, I think you can check that in OSF at least in some of the files

I haven't found any json file corresponding to this file.

phys2bids/tests/test_integration.py
```diff
@@ -233,3 +233,84 @@ def test_integration_multirun(skip_integration, multi_run_file):
...
+    if skip_integration:
+        pytest.skip('Skipping integration test')
+
+    test_path, test_filename = split(matlab_file)
+    test_chtrig = 1
+    conversion_path = join(test_path, 'code', 'conversion')
+
+    phys2bids(filename=test_filename, indir=test_path, outdir=test_path,
+              chtrig=test_chtrig, num_timepoints_expected=100, tr=1.2)
+
```
vinferrer, line 249: I have checked OSF and there isn't any json for ```test_2minRest_´´ so yeah we need @RayStick to tell us if the ``tr`` and ``num_timepoints_expected`` are correct

vinferrer:
Sry i didn't read you

> 
> 
> > You know, I think you can check that in OSF at least in some of the files
> 
> I haven't found any json file corresponding to this file.

RayStick:
I did not upload the test_2minRest.mat, I think it was @AyyagariA, so it is not easy for me to find the details (possible, but not easy).   

I would recommend using these files for tests:

First, look in test_data 
Second, if you can't find what you need in there try the files in _chicago/all_exports_. In that folder, I took time to export the same file in **loads** of possible formats, including _.mat_. Therefore, the file to use would be _chicago/all_exports/cut_file/CDB.mat._ There is just one json file for all these exports (_all_exports/CDB.json_) - not sure if that is useful but it shows you that the TR is 0.555 and num_tps is 1100. 

If those data files are not good for your tests, for whatever reason, I can try to track down details about test_2minRest.mat.

eurunuela:
Thank you @RayStick I think we can make it work with that data file. I'll push the changes tomorrow and hopefully we can have this merged in by the end of the month 💪🏻

eurunuela:
Just updated the tests to use the file _CBD.mat_ in _labchart/chicago/all_exports_. We should probably move it (or copy it) to the testing folder.

Let me know what you think of this PR @vinferrer @RayStick @smoia @tsalo

phys2bids/interfaces/mat.py
```diff
@@ -0,0 +1,72 @@
...
+    # Load MATLAB file into dictionary.
+    mat_dict = loadmat(filename)
+
+    # Convert data into 1d numpy array for easier indexing.
+    data = np.squeeze(np.asarray(mat_dict['data']))
+
+    # Extract number of channels and tick rate.
+    n_channels = len(mat_dict['titles'])
+    t_freq = mat_dict['tickrate'][0][0]
+
```
vinferrer, line 51: Here you are assuming t_freq is always in Hz. Are you sure this is the case always?

phys2bids/interfaces/mat.py
```diff
@@ -0,0 +1,72 @@
...
+    # Stores MATLAB data into lists.
+    timeseries = []
+    freq = [t_freq, ]
+    units = ['s', ]
+    names = ['time', ]
+
+    for ch in range(n_channels):
+        units.append(mat_dict['unittext'][int(mat_dict['unittextmap'][ch][0] - 1)].strip())
+        names.append(mat_dict['titles'][ch].strip())
+        freq.append(mat_dict['samplerate'][ch][0])
```
vinferrer, line 61: Same here, the sample rate could be in other frequency rather than Hz. I am talking from my experience with txt files.

vinferrer (changes requested):

phys2bids/interfaces/mat.py
eurunuela: Well, I didn't have any other info in the MATLAB files that we have in OSF. So I had to assume the frequencies are given in Hz (they definitely look like they're in Hz anyway).

Is this always the case @RayStick @smoia @tsalo ?

RayStick: I don't know if this is always the case (have not used this .mat formats before, generated from a labchart export) but I would assume it be Hz considering that is the units LabChart uses. I guess a .mat file that is not generated from LabChart could be different

vinferrer: Well I suppose that's something we have to specify then

phys2bids/interfaces/mat.py
```diff
@@ -0,0 +1,72 @@
...
+
+from phys2bids.physio_obj import BlueprintInput
+
+LGR = logging.getLogger(__name__)
+
+
+def populate_phys_input(filename, chtrig):
+    """
+    Populate object phys_input from MATLAB files.
+
```
vinferrer, line 19:
```suggestion
    Populate object phys_input from Labchart files in MATLAB format.

```

phys2bids/interfaces/mat.py
vinferrer: We should see if for example ``acq`` files can be in matlab format and add it to this function

vinferrer: imagen
Seems like acq does have matlab format. Therefore we can either leave it for another PR or add it in this one. In any of those case I would put a condition in the function to check that the ``.mat`` file is only in labchart format

eurunuela: What do you mean by "disregarding any other structure"? Do you only want labchart MATLAB files to be supported?

vinferrer: So if in this PR you only support the labchart structure, yes i want an if condition that checks something that you think is specific of labchart matlab files. In case you are gonna implement also de acq MATLAB files then accept only those two. Basically i want you to check that the ``.mat`` file people load is not some random MATLAB file

vinferrer: Mainly because this way we can raise a error

phys2bids/interfaces/mat.py
smoia: Actually, I think that the next lines deal with AcqKnowledge import as well?

phys2bids/tests/test_integration.py
```diff
@@ -233,3 +233,54 @@ def test_integration_multirun(skip_integration, multi_run_file):
     # for run in ['1', '2']:
     #     assert isfile(join(conversion_path, f'Test2_samefreq_TWOscans_{run}_trigger_time.png'))
     assert isfile(join(conversion_path, 'Test2_samefreq_TWOscans.png'))
+
+
+def test_integration_matlab(skip_integration, matlab_file):
+
+    if skip_integration:
```
smoia, line 240: Asking just because I'm seeing it here: why are we going for this approach when we could just tell pytest to run every test _but_ test_integration.py?

setup.cfg
```diff
@@ -41,20 +41,24 @@ doc =
     sphinx >=2.0
     sphinx-argparse
     sphinx_rtd_theme
+mat =
+    scipy>=1.5
```
smoia, line 45: See comment above about possibly adding h5py

phys2bids/interfaces/mat.py
```diff
@@ -0,0 +1,116 @@
...
+    chtrig is not a 0-based Python index - instead, it's human readable (i.e., 1-based).
+    This is handy because, when initialising the class, a new channel corresponding
+    to time is added at the beginning - that is already taken into account!
+
+    See Also
+    --------
+    physio_obj.BlueprintInput
+    """
+    # Load MATLAB file into dictionary.
+    mat_dict = loadmat(filename)
```
smoia, line 43: Ok I have quite a bad news here.
look at the `loadmat` Notes. Matlab files v7.3 aren't supported by it - instead, you need [h5py](https://www.h5py.org/) to [read those](https://stackoverflow.com/a/19340117). Alternatively you can break the code in case a matlab v7.3 file is given as input saying that we don't support it (yet).
In both cases, I would try this import and if it fails act in consequence (either trying using hdf5, or saying we don't support the format, or (better) both.

phys2bids/interfaces/mat.py
smoia: Ok, after a private chat with Eneko I got what's happening. Reverting this PR to draft!

smoia (commented):
Sorry, I went through this PR thinking it was ready to review. Here's just a comment about `loadmat`.
I'm reverting this in draft!

smoia:
@eurunuela @vinferrer, @62442katieb was asking if she can finish this PR. Are you working on it or not any more?

eurunuela:
I think @vinferrer was going to take care of it in the end.

62442katieb:
@eurunuela @smoia @vinferrer great! then I won't bother

eurunuela:
Let's make sure @vinferrer has done some work on it. I'm pretty sure he has but still.

smoia:
:rocket: Issue was released in [`2.3.0`](https://github.com/physiopy/phys2bids/releases/tag/2.3.0) :rocket:

PR #306, opened 2020-09-18
Auto-detect the trigger channel based on retrieved channel names

### LINKED ISSUE #257: Add 'auto' option for populate_phys_input chtrig parameter
Users may not know which channel is the trigger channel, so it would help them to be able to auto-detect the trigger channel and, in the absence of useable information, raise an error with the names of the available channels.
## Context / Motivation

For any user who doesn't know/remember what their trigger channel is, they have to load the data to a `BlueprintInput` with some random number set as the trigger channel, check the channel names, and then re-load the data with the correct trigger channel index. This should streamline the process in those cases. Plus, it would simplify the standard loading procedure in that there would only be one required argument to the function.

## Possible Implementation

- [ ] Create an alias list with "trig", "trigger", "TRIGGER", [...]
- [ ] raise a warning if the trigger channel's name does not contain any element in the alias list
- [ ] if the user didn't give a trigger channel, we look into the names and check if there is a match to any element of the alias list. If there is not, we raise an error.

vinferrer:
Closes #257 
auto-detect the trigger channel comparing ``ch_name`` list to a of possible trigger names
## Proposed Changes
  - [ ] Default chtrig is now 0 which will trigger the auto option
  - [ ]  Create an alias list with possible trigger names
  - [ ] raise a warning if the trigger channel's name does not contain any element in the alias list
  - [ ] if the user didn't give a trigger channel, we look into the names and check if there is a match to any element of the alias list. If there is not, we raise an error.


vinferrer:
Hello, @eurunuela, @tsalo, @smoia. I have implemented the auto option for the ``chtrig`` inside ``BlueprintInput`` class, because for me this makes more sense since it is the moment we start using the ``chtrig`` parameter. I left this as draft because i wan't you to corroborate that it makes sense to implement here the functionality the main reason is that independently of which interface we are running they all endup dropping the data in this class so no need for duplicities.
Also feel free to suggest new names for the chtrig alias names.

phys2bids/physio_obj.py
```diff
@@ -228,6 +228,21 @@ class BlueprintInput():
...
+                if trig_name in ch_name:
+                    trigger_idx = ch_name.index(trig_name)
+                    LGR.warning(f'Assigning chtrig to channel {trig_name}')
+                    name_not_found = False
+            if name_not_found:
+                raise Exception('No trigger channel was automaticly found. Exiting program')
+        else:
+            if ch_name[trigger_idx] not in trigger_names_list:
+                LGR.warning(f'Trigger channel name is not in our trigger channel name alias list.'
+                            f'Please make sure you are choosing the proper channel')
```
tsalo, line 245: No need for f-strings if there are no variables. Also needs a space after the period in the first line.

phys2bids/physio_obj.py
```diff
@@ -228,6 +228,21 @@ class BlueprintInput():
...
+        if trigger_idx == 0:
+            name_not_found = True
+            LGR.warning('User did not input chtrig. Trying to find in auto mode by name')
+            for trig_name in trigger_names_list:
+                if trig_name in ch_name:
+                    trigger_idx = ch_name.index(trig_name)
+                    LGR.warning(f'Assigning chtrig to channel {trig_name}')
+                    name_not_found = False
+            if name_not_found:
+                raise Exception('No trigger channel was automaticly found. Exiting program')
```
tsalo, line 241:
```suggestion
                raise Exception('No trigger channel was automatically found. Exiting program.')
```

phys2bids/physio_obj.py
```diff
@@ -228,6 +228,21 @@ class BlueprintInput():
     def __init__(self, timeseries, freq, ch_name, units, trigger_idx,
                  num_timepoints_found=None, thr=None, time_offset=0):
         """Initialise BlueprintInput (see class docstring)."""
+        trigger_names_list = ["trig", "trigger", "TRIGGER", "Trigger"]
+        if trigger_idx == 0:
+            name_not_found = True
+            LGR.warning('User did not input chtrig. Trying to find in auto mode by name')
```
tsalo, line 234: Could you add periods to the ends of the strings too? For consistency- since there are multiple sentences and all of the other sentences have punctuation.

phys2bids/physio_obj.py
```diff
@@ -228,6 +228,21 @@ class BlueprintInput():
...
+            for trig_name in trigger_names_list:
+                if trig_name in ch_name:
+                    trigger_idx = ch_name.index(trig_name)
```
tsalo, line 237: Do we want a complete string match or do we want to support partial matches? If the latter, then I think a regular expression would probably be easier here. If the former, then I'd switch the comparison (loop through `ch_name` and see if each name is "in" `trigger_names_list`, probably with a list comprehension). We also want to raise an exception or at least a warning if multiple channel names match the criteria.

tsalo:
I'll leave the discussion of the proper index for the trigger channel (i.e., 0 or 1) and where that switch happens to others, but I had some minor code edits/comments. Thanks for this!

phys2bids/cli/run.py
```diff
@@ -77,8 +77,8 @@ def _get_parser():
                           type=int,
                           help='The column number of the trigger channel. '
                                'Channel numbering starts with 1. '
```
smoia, line 79: We need a line or two that explains what happens when chtrig is zero!

phys2bids/interfaces/txt.py
```diff
@@ -76,7 +76,7 @@ def process_labchart(channel_list, chtrig, header=[]):
     channel_list: list
         list with channels only
     chtrig : int
-        index of trigger channel
+        index of trigger channel, starting in 1 for human readability
```
smoia, line 79: And at this point, should we state that their count starts in 1, but zero is a correct input - and what it does?
(Here and in every other docstring)

phys2bids/phys2bids.py
```diff
@@ -191,7 +191,7 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     # Check options to make them internally coherent pt. II
     # #!# This can probably be done while parsing?
     indir = utils.check_input_dir(indir)
-    if chtrig < 1:
+    if chtrig < 0:
         raise Exception('Wrong trigger channel. Channel indexing starts with 1!')
```
smoia, line 195: This message has to be updated accordingly (starts with 1, or it's zero for automatic search)

phys2bids/physio_obj.py
```diff
@@ -228,6 +228,21 @@ class BlueprintInput():
     def __init__(self, timeseries, freq, ch_name, units, trigger_idx,
                  num_timepoints_found=None, thr=None, time_offset=0):
         """Initialise BlueprintInput (see class docstring)."""
+        trigger_names_list = ["trig", "trigger", "TRIGGER", "Trigger"]
```
smoia, line 231: This could become a constant, initialised at the beginning of the file - it would also be easier to find if we need to update it.

Another thing that could be done is to:
- have only lowercases/uppercases (so you don't have to specify all the variations of `trigger`, for instance, like we're doing for the units of measure),
- order the items from the longest to the shortest (so that first `trigger` is searched, then `trigg`, ... then, if we want, `t`),
- implement a partial match (like @tsalo is suggesting further).

In case we only use the lowercase or uppercase, we can rename the channel `trigger` (since that is the suggested BIDS name).

This would reduce the length of the list and support more possibilities.

phys2bids/physio_obj.py
smoia: I agree! See my comment above.

phys2bids/physio_obj.py
```diff
@@ -228,6 +228,21 @@ class BlueprintInput():
...
+        if trigger_idx == 0:
+            name_not_found = True
+            LGR.warning('User did not input chtrig. Trying to find in auto mode by name')
+            for trig_name in trigger_names_list:
+                if trig_name in ch_name:
+                    trigger_idx = ch_name.index(trig_name)
+                    LGR.warning(f'Assigning chtrig to channel {trig_name}')
```
smoia, line 238:
```suggestion
                    LGR.warning(f'{trig_name} selected as trigger channel')
```

phys2bids/physio_obj.py
smoia:
```suggestion
                raise Exception('No trigger channel was automatically found. Please run phys2bids specifying the -chtrig argument')
```

smoia (changes requested):
I can't wait to see this implemented and working!

However, I have few comments and changes to ask (see comments).

There's another thing: I like the fact that this functionality is implemented in the object methods, but I don't know if it's actually good to have it there, and in any case I wouldn't put it in the initialisation of the object.

The reason for which I'm not sure it's good to have it in the object is that I think we're using the trigger channel to extract the time array when we deal with acq files (in the respective interface script), and that's a step that takes place before the initialisation of the object.
Maybe a better place would be the (refactored) interfaces script/folder, with this step taking place before the input processing. It's not vital though, we can just specify the change in the initialisation of acq files.

If we want to leave it as an object method (and I see why), I would make it a separate method and call it in case chtrig is zero, even after the initialisation of the object. It just means updating its properties.

phys2bids/physio_obj.py
```diff
@@ -228,6 +228,21 @@ class BlueprintInput():
...
+                    name_not_found = False
+            if name_not_found:
+                raise Exception('No trigger channel was automatically found. Exiting program.')
```
smoia, line 241:
```suggestion
                raise Exception('No trigger channel was automatically found. Please run phys2bids specifying the -chtrig argument.')
```

eurunuela:
Hi @vinferrer , thanks a lot for this! I agree with the comments from @tsalo and @smoia . My other concern would be the linting not passing the test. I think once the comments are added we can merge it. Thanks again!

vinferrer:
Okay i have 

> I can't wait to see this implemented and working!
> 
> However, I have few comments and changes to ask (see comments).
> 
> There's another thing: I like the fact that this functionality is implemented in the object methods, but I don't know if it's actually good to have it there, and in any case I wouldn't put it in the initialisation of the object.
> 
> The reason for which I'm not sure it's good to have it in the object is that I think we're using the trigger channel to extract the time array when we deal with acq files (in the respective interface script), and that's a step that takes place before the initialisation of the object.
> Maybe a better place would be the (refactored) interfaces script/folder, with this step taking place before the input processing. It's not vital though, we can just specify the change in the initialisation of acq files.
> 
> If we want to leave it as an object method (and I see why), I would make it a separate method and call it in case chtrig is zero, even after the initialisation of the object. It just means updating its properties.

I see what you mean about acq files. Is it critical that you use the chtrig channel to obtain the frequency and the time index?

vinferrer:
So guys I have implemented the partial match thingy as you can see, if you don't want me to have this in physio_obj.py we could put it as a utils function that is used in each of the interfaces.py script but i think that would make more duplicities. Is there a way we can avoid using chtrig in acq.py?

eurunuela:
> So guys I have implemented the partial match thingy as you can see, if you don't want me to have this in physio_obj.py we could put it as a utils function that is used in each of the interfaces.py script but i think that would make more duplicities. Is there a way we can avoid using chtrig in acq.py?

I agree that not using `chtrig` in `acq.py` would make the code cleaer. Any idea @smoia ? I believe that piece of code was yours.

vinferrer:
Now there is a new method for this functionality that is called in the inicialization

smoia:
It's not strictly mandatory to use chtrig to read acq files. The reason for which it's used is due to the fact that the time array in acq files has to be extracted from a channel, and in order to avoid issues with other functions that come later (like getting the right starting time of the sequence), I extract it from the trigger channel.

I would need to integrate (not-so-)little changes to different parts of the code in order to be able not to use the trigger channel in the acq reads, like checking that the time channel and the trigger channel have the same frequency and eventually interpolate them.

vinferrer:
So we need to discuss the ``chtrig`` default and if we can avoid using ``chtrig`` in ``acq.py``

smoia:
Indeed! I'm not against not using it in the file loading, but we need a bit more work to be able to make it independent.
Do you want to wait for this to happen? That's another PR that might take time.

vinferrer:
> Indeed! I'm not against not using it in the file loading, but we need a bit more work to be able to make it independent.
> Do you want to wait for this to happen? That's another PR that might take time.

Okay i just need to understand why you use the chtrig in ``acq.py`` in the case of the frequency we could always take the maximun frequency which would correspond to the trigger mostly. I am pretty sure we can look for ways to avoid using ``chtrig`` and there would be no need of opening a new PR

vinferrer:
@smoia I implemented a solution in ``acq.py`` that avoids using chtrig I think is the most ideal way creating the time channel

smoia:
@vinferrer the problem is not `acq.py` itself: we could simply use channel 0 to extract the time - much easier and computationally inexpensive!
The issue is the implementation of the starting time discovery in `physio_obj.py` (around line 473).
If the trigger channel and the time channel don't have the same frequency they won't have the same number of elements,  and the starting time will not be picked up correctly.
If you want to solve the problem, you could do an interpolation of the time channel at the frequency of the trigger channel before the selection of the starting time. If you do that, please open a different PR with the changes to acq.py and physio_obj.py .

smoia:
@vinferrer the problem is not `acq.py` itself: we could simply use channel 0 to extract the time - much easier and computationally inexpensive!
The issue is the implementation of the starting time discovery in `physio_obj.py` (around line 473).
If the trigger channel and the time channel don't have the same frequency they won't have the same number of elements,  and the starting time will not be picked up correctly.
If you want to solve the problem, you could do an interpolation of the time channel at the frequency of the trigger channel before the selection of the starting time. If you do that, please open a different PR with the changes to acq.py and physio_obj.py .

smoia:
@vinferrer the problem is not `acq.py` itself: we could simply use channel 0 to extract the time - much easier and computationally inexpensive!
The issue is the implementation of the starting time discovery in `physio_obj.py` (around line 473).
If the trigger channel and the time channel don't have the same frequency they won't have the same number of elements,  and the starting time will not be picked up correctly.
If you want to solve the problem, you could do an interpolation of the time channel at the frequency of the trigger channel before the selection of the starting time. If you do that, please open a different PR with the changes to acq.py and physio_obj.py .

vinferrer:
> 
> 
> @vinferrer the problem is not `acq.py` itself: we could simply use channel 0 to extract the time - much easier and computationally inexpensive!
> The issue is the implementation of the starting time discovery in `physio_obj.py` (around line 473).
> If the trigger channel and the time channel don't have the same frequency they won't have the same number of elements, and the starting time will not be picked up correctly.
> If you want to solve the problem, you could do an interpolation of the time channel at the frequency of the trigger channel before the selection of the starting time. If you do that, please open a different PR with the changes to acq.py and physio_obj.py .

I see what you mean. However, I think you are not considering that it would be very weird that the trigger channel frequency is not the maximum one in the file. Think about it, how many physiological files have you encontered where there is a channel with higher frequency than the trigger channel? I would say zero, though I am not an expert. Is this case normal?

smoia:
You might be right, but as a general approach, if we want to obtain more generalisation in our tools, then thinking about the case that is least probable to have to be changed again is a better way to go than a solution that might be true, but we're not sure about.

It's true that in our case (BCBL) the trigger will be the channel with highest frequency, and possibly in the majority of cases it will be in the highest frequency group, but think, for instance, about @tsalo 's setting at FIU, described in #219 , where they don't collect the trigger continuously (but rather just beginning and end of run). In that case your sampling of the trigger can be much, much lower.

Since the interpolation of the time channel is a more generalisable approach to solve this issue, I'd rather change that part of code. I can do it later in time, but if you're in a hurry and want to do it now, that's a more generalisable solution (it shouldn't be more than three lines of code) -  to merge in another PR though.

vinferrer:
> It's true that in our case (BCBL) the trigger will be the channel with highest frequency, and possibly in the majority of cases it will be in the highest frequency group, but think, for instance, about @tsalo 's setting at FIU, described in #219 , where they don't collect the trigger continuously (but rather just beginning and end of run). In that case your sampling of the trigger can be much, much lower.

Okey, we need to discuss this, however if we speak about #219 it would be a horrible idea to create a time channel with an uncontinous trigger, because it wouldn't be representative of the rest, one reason to change your code. The setting in FIU is a weird exception that we don't cover yet and we would have to change much more than ``acq.py```. Regarding the interpolation of the time channel, I think I already create the time channel with ``np.linespace``` in ``txt.py`` but again taking the maximum frequency as base to create it. I could adapt it to ``acq.py```. If you want to discuss this futher we can do a quick meeting tomorrow.

tsalo:
For clarification- FIU's trigger setup _is_ continuous. It's just a block-ish binary signal.

vinferrer:
Oh my bad, i missunderstood

vinferrer:
Well then, I will leave this as Draft until the ``chtrig``  in ``acq.py`` is modified

smoia:
Shouldn't we wait for #308 to be merged?

vinferrer:
Hey peole, I think this is ready for review @tsalo @smoia

smoia (changes requested):
Some changes suggested, mainly language related, but we're getting close.

I have an alternative implementation for your `re` search. Compare the two solutions:
```
import re

TRIGGER_NAMES = ["trig", "trigger"]

# Adding some possible names. I would expect to find a match to ch_name[1, 3, 4, 6, 7, 8, 9]
ch_name = ['trg', 'trig', 'trigg', 'TRIGGER', 'Trig', 'TRIGGAH', 'TRIG 1000', 'TRIG1000', 'trig_channel', 'trig channel', 'strigose', 'gas_no_trigger']

# current solution returns all but ch_name[0]
results = [re.search('|'.join(TRIGGER_NAMES), case, re.IGNORECASE) for case in ch_name]
results

# alternative formulation still returns ch_names[2, 11], which is wrong, but doesn't return ch_names[5, 10].
results = [re.search('|'.join(re.split('(\W+|\d|_)', case)), '§'.join(TRIGGER_NAMES), re.IGNORECASE) for case in ch_name]
results
```

This solution is a bit more specific, and could be generalised to a matching pattern for more channel types (e.g. gas, belts, cardiac, ...), should we want to offer renaming - granted, not so much better, which is why I don't love it. Maybe someone else comes up with a better solution.

In any case, can you describe a bit further the solution you pick up to find the matching `ch_names`? It's not so easy to get the idea behind it straight away.

Also note that for the current solution, at the moment the only thing you need to compare is 'trig', given that 'trigg', 'trigger', etc. all return a match to 'trig'. If you decide to keep this one, please change `TRIGGER_NAMES` to `TRIGGER_PATTERN = ['trig']`, since it would be more appropriate - and give you the same result.

phys2bids/physio_obj.py
```diff
@@ -238,6 +241,12 @@ def __init__(self, timeseries, freq, ch_name, units, trigger_idx,
         self.num_timepoints_found = num_timepoints_found
         self.thr = thr
         self.time_offset = time_offset
+        if trigger_idx == 0:
+            self.auto_trigger_selection()
+        else:
+            if ch_name[trigger_idx] not in trigger_names_list:
+                LGR.warning('Trigger channel name is not in our trigger channel name alias list. '
```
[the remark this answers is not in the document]
vinferrer: it doesn't make any damage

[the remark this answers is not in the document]
smoia: Well, warnings mean that something peculiar is happening. At least I would consider downgrading this message to an `info` level.

phys2bids/tests/test_physio_obj.py
```diff
@@ -259,3 +258,39 @@ def test_BlueprintOutput():
...
+    test_time = np.array([0, 1, 2, 3, 4])
+    test_trigger = np.array([0, 1, 2, 3, 4])
+    test_half = np.array([0, 1, 2])
+    test_twice = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+    test_timeseries = [test_time, test_trigger, test_half, test_twice]
+    test_freq = [1, 1, 0.5, 2]
+    test_chn_name = ['time', 'trigger', 'half', 'twice']
+    test_units = ['s', 'V', 'V', 'V']
+
+    # test when trigger is not the name of the channel
```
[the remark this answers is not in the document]
vinferrer: WHY?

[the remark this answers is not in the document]
smoia: Because I was equally suggesting to remove the related warning.

vinferrer:
> Some changes suggested, mainly language related, but we're getting close.
> 
> I have an alternative implementation for your `re` search. Compare the two solutions:
> 
> ```
> import re
> 
> TRIGGER_NAMES = ["trig", "trigger"]
> 
> # Adding some possible names. I would expect to find a match to ch_name[1, 3, 4, 6, 7, 8, 9]
> ch_name = ['trg', 'trig', 'trigg', 'TRIGGER', 'Trig', 'TRIGGAH', 'TRIG 1000', 'TRIG1000', 'trig_channel', 'trig channel', 'strigose', 'gas_no_trigger']
> 
> # current solution returns all but ch_name[0]
> results = [re.search('|'.join(TRIGGER_NAMES), case, re.IGNORECASE) for case in ch_name]
> results
> 
> # alternative formulation still returns ch_names[2, 11], which is wrong, but doesn't return ch_names[5, 10].
> results = [re.search('|'.join(re.split('(\W+|\d|_)', case)), '§'.join(TRIGGER_NAMES), re.IGNORECASE) for case in ch_name]
> results
> ```
> 
> This solution is a bit more specific, and could be generalised to a matching pattern for more channel types (e.g. gas, belts, cardiac, ...), should we want to offer renaming - granted, not so much better, which is why I don't love it. Maybe someone else comes up with a better solution.
> 
> In any case, can you describe a bit further the solution you pick up to find the matching `ch_names`? It's not so easy to get the idea behind it straight away.
> 
> Also note that for the current solution, at the moment the only thing you need to compare is 'trig', given that 'trigg', 'trigger', etc. all return a match to 'trig'. If you decide to keep this one, please change `TRIGGER_NAMES` to `TRIGGER_PATTERN = ['trig']`, since it would be more appropriate - and give you the same result.

According to my test the first expression gives these indexes:
 [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
and the second:
 [1, 2, 3, 4, 6, 7, 8, 9, 11]

it's true that it is more specific, however there is no way to assure this method will work, that's why i put the raises

vinferrer:
and by the way, your expression gives this error:
image

vinferrer:
> results = [re.search('|'.join(re.split('(\W+|\d|_)', case)), '§'.join(TRIGGER_NAMES), re.IGNORECASE) for case in ch_name]
> results

Problem, this expresion detects the channel CO2 as trigger!!!: look at the fail in the unit test

phys2bids/phys2bids.py
```diff
@@ -191,8 +191,11 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     # Check options to make them internally coherent pt. II
     # #!# This can probably be done while parsing?
     indir = utils.check_input_dir(indir)
-    if chtrig < 1:
-        raise Exception('Wrong trigger channel. Channel indexing starts with 1!')
+    if chtrig <= 0:
+        raise Exception('Wrong trigger channel. Channel indexing cannot be negative!')
+    elif chtrig == 0:
+        LGR.warning('No trigger channel specified. phys2bids will run an automatic search based '
```
eurunuela, line 197: We should add a test that checks for this warning.

phys2bids/physio_obj.py
```diff
@@ -238,6 +241,12 @@ def __init__(self, timeseries, freq, ch_name, units, trigger_idx,
         self.num_timepoints_found = num_timepoints_found
         self.thr = thr
         self.time_offset = time_offset
+        if trigger_idx == 0:
+            self.auto_trigger_selection()
+        else:
+            if ch_name[trigger_idx] not in TRIGGER_NAMES:
+                LGR.warning('Trigger channel name is not in our trigger channel name alias list. '
+                            'Please make sure you are choosing the proper channel')
```
eurunuela, line 249:
```suggestion
                            'Please make sure you choose the proper channel.')
```

phys2bids/physio_obj.py
```diff
@@ -542,6 +551,39 @@ def print_info(self, filename):
...
+        More than one possible trigger channel was automatically found.
+        Exception
+        No trigger channel automatically found
+        Notes
+        -----
+        Outcome:
+            trigger_idx : int
+                Automatically retrieved trigger index
+        """
+        LGR.warning('Running automatic trigger detection.')
```
eurunuela, line 573: I don't think this should be a warning but rather an `INFO`. phys2bids will have already warned the user after the parser. Now, I see this function as a feature in phys2bids, not as a function that tries to fix an issue or a bad argument the user has introduced.

phys2bids/physio_obj.py
```diff
@@ -542,6 +551,39 @@ def print_info(self, filename):
...
+        results = [re.search('|'.join(TRIGGER_NAMES),
+                             case, re.IGNORECASE) for case in self.ch_name]
+        indexes = [i for i, v in enumerate(results) if v]
+        if len(indexes) == 1:
+            self.trigger_idx = indexes[0]
+            LGR.warning(f'{self.ch_name[self.trigger_idx]} selected as trigger channel')
```
eurunuela, line 579: Again, this is `INFO` we'd be giving the user, not a warning message.

eurunuela (changes requested):
I think we should have `INFO` messages when information is given to the user and not warnings. I would only keep the warning that warns the user about phy2bids using the automatic trigger channel option. The rest should be `INFO` in my opinion.

The changes look good to me otherwise, thank you @vinferrer !

phys2bids/phys2bids.py
vinferrer: You mean for the negative warning?

eurunuela: What do you mean by negative warning?

I didn't see a test checking that this warning message is shown when `chtrig==0` but I might've missed it.

vinferrer:
```
    # test when trigger is 0 and that the trigger channel is recognized by name:
    test_chtrig = 0
    phys_in = po.BlueprintInput(test_timeseries, test_freq, test_chn_name,
                                test_units, test_chtrig)
    assert phys_in.trigger_idx == 1
```
Look in ``test_physio_obj``

eurunuela: I see, thanks!

eurunuela (approved):
LGTM! Thanks!

vinferrer:
@eurunuela, I have changed the regular expression in the end to @smoia solution but deleting first the numbers. Could you review this little bit?

vinferrer:
@smoia  could we close this matter?

phys2bids/physio_obj.py
```diff
@@ -543,6 +552,40 @@ def print_info(self, filename):
...
+        ------
+        Exception
+        More than one possible trigger channel was automatically found.
+        Exception
+        No trigger channel automatically found
+        Notes
+        -----
+        Outcome:
+            trigger_idx : int
+                Automatically retrieved trigger index
```
tsalo, line 572: There are some spacing issues in this docstring. The main thing is the lack of empty lines between docstring sections.

phys2bids/physio_obj.py
```diff
@@ -543,6 +552,40 @@ def print_info(self, filename):
 
         LGR.info(info)
 
+    def auto_trigger_selection(self):
```
tsalo, line 555: I think we need two empty lines here.

tsalo (changes requested):
This looks great. TBH I'd prefer it if `chtrig` accepted a special string value (e.g., `"auto"`) when you want automatic trigger channel detection. However, I think it is totally cool if 0 is used, if folks don't feel the same way. If so, then the main things I noticed are minor issues with the function docstring.

phys2bids/physio_obj.py
vinferrer: > This looks great. TBH I'd prefer it if `chtrig` accepted a special string value (e.g., `"auto"`) when you want automatic trigger channel detection. However, I think it is totally cool if 0 is used, if folks don't feel the same way. If so, then the main things I noticed are minor issues with the function docstring.

Does someone else have an opinion on the auto option?

vinferrer:
> This looks great. TBH I'd prefer it if `chtrig` accepted a special string value (e.g., `"auto"`) when you want automatic trigger channel detection. However, I think it is totally cool if 0 is used, if folks don't feel the same way. If so, then the main things I noticed are minor issues with the function docstring.

Does someone else have an opinion on the auto option?

smoia:
@tsalo at the moment the CLI accepts only integers for the `chtrig` option (the one used here). Wouldn't it be a problem to have have a "auto" string as default instead?

tsalo:
You're right. It would require a couple of changes to the CLI, although we have a good template for that kind of argument in `tedana` now, with `--tedpca`. See [here](https://github.com/ME-ICA/tedana/blob/14be8520da0ba577c130c56d0835c55321cd0aed/tedana/workflows/tedana.py#L99-L110).

If it seems like too much work, or otherwise just not a good idea, what about `-1`? Most Python users would think `0` would be a standard value, while `-1` is a common "special value".

vinferrer:
Okay, actually I do have a very strong opinion about using ``auto`` or -1 as the default option, think that the check is done after reading the file, all our ``io.py`` are designed to use ``chtrig`` as a number and there is actually an error designed to pop up when ``chtrig`` is negative, so I would stay with the 0 as default option

tsalo (approved):
LGTM!

phys2bids/physio_obj.py
```diff
@@ -239,6 +242,12 @@ def __init__(self, timeseries, freq, ch_name, units, trigger_idx,
         self.num_timepoints_found = deepcopy(num_timepoints_found)
         self.thr = deepcopy(thr)
         self.time_offset = deepcopy(time_offset)
+        if trigger_idx == 0:
+            self.auto_trigger_selection()
+        else:
+            if ch_name[trigger_idx] not in TRIGGER_NAMES:
+                LGR.warning('Trigger channel name is not in our trigger channel name alias list. '
```
smoia, line 249:
```suggestion
                LGR.info('Trigger channel name is not in our trigger channel name alias list. '
```

phys2bids/physio_obj.py
```diff
@@ -543,6 +552,43 @@ def print_info(self, filename):
...
+        """
+        LGR.info('Running automatic trigger detection.')
+        no_numbers = [re.sub(r'\d+', '', x) for x in self.ch_name]
+        results = [re.search('|'.join(re.split(r'(\W)', case)),
+                             '§'.join(TRIGGER_NAMES), re.IGNORECASE) for case in no_numbers]
+        indexes = [i for i, v in enumerate(results) if v]
+        if len(indexes) == 1:
+            self.trigger_idx = indexes[0]
+            LGR.info(f'{self.ch_name[self.trigger_idx]} selected as trigger channel')
+        if len(indexes) > 1:
```
smoia, line 585:
```suggestion
        elif len(indexes) > 1:
```

phys2bids/physio_obj.py
```diff
@@ -543,6 +552,43 @@ def print_info(self, filename):
...
+            raise Exception('More than one possible trigger channel was automatically found. '
+                            'Please run phys2bids specifying the -chtrig argument.')
+        if len(indexes) < 1:
```
smoia, line 588:
```suggestion
        else:
```

smoia:
I checked again the regex comparison, and came up with another alternative solution that doesn't have the issue of matching wrong channel names due to empty strings in the channel name split.
Granted, it's less elegant because it uses more lines of code, but it's a bit more specific.
I opened a PR to your branch, [here](https://github.com/vinferrer/phys2bids/pull/8) if @tsalo or @eurunuela want to have a look at it.

vinferrer:
Well I think we are ready to merge this

smoia (approved):
LGTM, but I'm biased!
@tsalo @eurunuela do you want to give a second look? If so, the last one can merge this in!

tsalo (approved):
LGTM

smoia:
:rocket: PR was released in [`2.4.0`](https://github.com/physiopy/phys2bids/releases/tag/2.4.0) :rocket:

PR #308, opened 2020-10-02
Make trigger-based time-offset correction more generalisable (resample time before comparing it to trigger)

smoia:
Closes none, but addresses #306 

## Proposed Changes

  - `interfaces.acq` initialisation is now independent from the trigger
  - The trigger channel number is still passed to the file reading functions for objects initialisation
  - If time and trigger channel don't have the same length, time is (locally) resampled to match the trigger length


phys2bids/interfaces/acq.py
```diff
@@ -42,8 +39,8 @@ def populate_phys_input(filename, chtrig):
         warnings.filterwarnings('ignore', category=DeprecationWarning)
         data = read_file(filename).channels
 
-    freq = [data[chtrig - 1].samples_per_second, ]
-    timeseries = [data[chtrig - 1].time_index, ]
+    freq = [data[0].samples_per_second, ]
+    timeseries = [data[0].time_index, ]
```
vinferrer, line 43:
```suggestion
    frequencies = [x.samples_per_second for x in data]
    max_index = [i for i, j in enumerate(frequencies) if j == max(frequencies)][0]
    freq = [data[max_index].samples_per_second, ]
    interval = 1/freq[0]
    duration = (timeseries[max_index].shape[max_index] + 1) * interval
    timeseries = [np.ogrid[0:duration:interval][:-1], ]  # create time channel
```
This will make your frequency time and time channel independent from any index because you will always get the biggest one

phys2bids/physio_obj.py
```diff
@@ -456,7 +456,14 @@ def check_trigger_amount(self, thr=None, num_timepoints_expected=0, tr=0):
         # Use the trigger channel to find the TRs,
         # comparing it to a given threshold.
         trigger = self.timeseries[self.trigger_idx]
+        time = self.timeseries[0]
         LGR.info(f'The trigger is in channel {self.trigger_idx}')
+        # Check that trigger and time channels have the same length.
+        # If not, resample time to the length of the trigger
+        if len(time) != len(trigger):
+            time = np.linspace(time[0], time[-1], len(trigger))
```
vinferrer, line 464: you wouldn't need this
```suggestion
```

phys2bids/physio_obj.py
```diff
@@ -470,7 +477,7 @@ def check_trigger_amount(self, thr=None, num_timepoints_expected=0, tr=0):
         else:
             LGR.info(f'The number of timepoints found with the manual threshold of {thr:.4f} '
                      f'is {num_timepoints_found}')
-        time_offset = self.timeseries[0][timepoints.argmax()]
+        time_offset = time[timepoints.argmax()]
```
vinferrer, line 480:
```suggestion
        time_offset = self.timeseries[0][timepoints.argmax()]
```

smoia:
@vinferrer I understand your suggested changes, however please consider that:
- Even if you set your time to be taken from the channel with the highest frequency, you still need to check at a later point that its size matches with your trigger. It's not a given that your trigger channel is the one with the highest frequency. For instance, if for any reason a user decides that a PPG should be sampled at 1000 Hz, while they're happy with a 40 Hz trigger sampling, the suggested change would not provide the right time offset.
- The solution I'm proposing is very condensed in terms of code (in fact, it's two lines of code more that create an `if` statement, that might not be run). The suggested changes are more than twice the size, with two `for` loops. It would still need the original `if`statement (see point above). It's a lot of computation more for a feature I'm not sure is needed.

I wouldn't accept the suggested changes, but if a second reviewer (@eurunuela? @tsalo?) agrees on them, I'll change the code.

vinferrer:
You have a good point there, I think we can go with your solution

vinferrer (approved):
LGTM!! however, we need one more reviewer

phys2bids/physio_obj.py
eurunuela, line 464: I believe these two lines here are not tested. Could you open an issue for this please? Or add a quick assert to the  test. Thanks!

eurunuela (commented):
I'm happy with the changes, just want you to elaborate a bit on two of the lines as I may have missed the discussion here.

smoia:
I added some asserts and a full test for this PR. I don't understand why codecov is unhappy though. @eurunuela let me know if there is something else I can do!

eurunuela (approved):
LGTM! Thank you @smoia !

smoia:
@vinferrer if you're ok with the updates, you can approve and merge in (and then we start again with #306)!

phys2bids/interfaces/acq.py
```diff
@@ -12,17 +12,14 @@
 LGR = logging.getLogger(__name__)
 
 
-def populate_phys_input(filename, chtrig):
+def populate_phys_input(filename, chtrig=0):
```
tsalo, line 15: If `chtrig` isn't being used anymore (and based on the docstring change that's what I'm inferring), then why keep the argument?

phys2bids/physio_obj.py
```diff
@@ -504,7 +512,7 @@ def check_trigger_amount(self, thr=None, num_timepoints_expected=0, tr=0):
                         'were not provided.')
         self.thr = thr
         self.time_offset = time_offset
-        self.timeseries[0] -= time_offset
+        self.timeseries[0] = self.timeseries[0] - time_offset
```
tsalo, line 515: Why have this change?

phys2bids/tests/test_physio_obj.py
```diff
@@ -89,7 +91,7 @@ def test_BlueprintInput():
     test_time = np.array([0, 1, 1, 2, 3, 5, 8, 13])
     test_trigger = np.array([0, 1, 0, 0, 0, 0, 0, 0])
     test_chocolate = np.array([1, 0, 0, 1, 0, 0, 1, 0])
-    test_timeseries = [test_time, test_trigger, test_chocolate]
+    test_timeseries = [deepcopy(test_time), test_trigger, test_chocolate]
```
tsalo, line 94: Why deepcopy this array?

phys2bids/physio_obj.py
```diff
@@ -456,7 +457,14 @@ def check_trigger_amount(self, thr=None, num_timepoints_expected=0, tr=0):
         # Use the trigger channel to find the TRs,
         # comparing it to a given threshold.
         trigger = self.timeseries[self.trigger_idx]
+        time = deepcopy(self.timeseries[0])
         LGR.info(f'The trigger is in channel {self.trigger_idx}')
+        # Check that trigger and time channels have the same length.
+        # If not, resample time to the length of the trigger
+        if len(time) != len(trigger):
+            time = np.linspace(time[0], time[-1], len(trigger))
```
tsalo, line 465: Is this worth a warning? Given how this interacts with #306, and depending on how uncommon this is, I think that having time not match the trigger might indicate that the automatic detection of the trigger channel made a mistake.

phys2bids/interfaces/acq.py
vinferrer: basically chtrig is called at the end of the function by ``populate_phys_auto``

tsalo: Ah, thank you. It needs to stay in the docstring then.

phys2bids/interfaces/acq.py
```diff
@@ -12,17 +12,14 @@
...
     """
     Populate object phys_input from acq files.
 
     Parameters
     ----------
     filename: str
         path to the txt labchart file
-    chtrig : int
-        index of trigger channel.
-        !!! ATTENTION: IT'S MEANT TO REPRESENT AN INDEX STARTING FROM 1 !!!
```
tsalo, line 25: Can you revert this? Perhaps something like

```python
    chtrig : int, optional
          index of trigger channel. Default is 0.
```

phys2bids/physio_obj.py
smoia: It's [this commit](https://github.com/physiopy/phys2bids/pull/308/commits/2d3e378b7f1ecf18b0ea482008ad3b9837bcc924). Basically, while testing I got some dtype errors of impossible subtraction between dtype X and dtype Y with some peculiar cases of `self.timeseries[0]` and `time_offset`. I read that an easy way to avoid them is using a complete subtraction statement rather than the previous version. It's a more global solution (despite possibly less elegant).

phys2bids/tests/test_physio_obj.py
smoia: If we don't make a deep copy, it's just a shallow reference, so `test_time` changes when `blueprint_in.check_trigger_amount()` is called, and we wouldn't be able to set up the new asserts.

phys2bids/physio_obj.py
tsalo: That makes sense. I've had issues with in-place subtraction in the past too. Thanks for the explanation!

phys2bids/tests/test_physio_obj.py
tsalo: I think that's a bug on `BlueprintInput`'s end rather than something that the test should work around.

vinferrer: Actually this is related to the effect of shallow copies vs deepcopies: Basically, a shallow copy does copy the main object but references child objects to the original. Therefore if you modify the child object of the original, it also gets modified in the copy. If you deepcopy, you create a completely independent object (that consumes more memory). More info:
https://realpython.com/copying-python-objects/

vinferrer: It is an undesired behavior but not a bug, just how python works

smoia: Yup, @vinferrer is right. For our scopes, I don't think it's an undesired behaviour either.
However, if @tsalo you think there's a better solution to approach this, we can open an issue to change this - I wouldn't block this PR for this, though.

tsalo: When I say it is a bug, I mean that a given function/class should not to modify its inputs directly unless that behavior is desired and documented. A good example of this is [`nilearn.image.get_data()`](https://nilearn.github.io/modules/generated/nilearn.image.get_data.html#nilearn.image.get_data), which is an alternative to nibabel's Nifti1Image's `get_fdata()`. Unlike nibabel's method, nilearn returns an img object's numpy array directly instead of a copy. Even with that warning, there are still unexpected behaviors like https://github.com/nilearn/nilearn/issues/2548, so I'd lean against it. That issue I linked was only a bug _after_ they switched from nibabel's function (return a copy) to nilearn's function (return a reference).

On the other hand, when the behavior isn't documented, we can get issues like https://github.com/nilearn/nilearn/issues/2281 and https://github.com/nilearn/nilearn/issues/2475. 

There are valid reasons to not make copies of all of your inputs within a function of course, like memory. I can't speak to the exact memory requirements of a BlueprintInput, but I imagine they could be quite large given the right circumstances. If you don't want to make a copy _within_ the BlueprintInput, the we should probably add something to the docstring about this behavior. `nilearn.image.get_data()`'s docstring could serve as a good template.

This is a minor thing, but since this numpy array is just an array of integers or floats, and not objects, shallow copying should be fine. You could use `test_time.copy()` without issue. The problem is that `BlueprintInput` doesn't make a copy of the array at all and operates on the original array directly.

smoia: @tsalo now it's clear. Could you open an issue for this? I'm going to change the deepcopy to copy.

tsalo: I've opened #314 (what luck!). I'll resolve this thread, but fixing the bug will involve dropping the copy in the test in a later PR. Thanks!

vinferrer:
Should we merge then?

vinferrer:
Or do you want us to re-review?

tsalo (approved):
LGTM.

PR #312, opened 2020-10-09
add format parameter to the logger

### LINKED ISSUE #196: Making terminal output easier to read
## Summary

- This is a very minor issue.  There is quite a lot of text outputted to the terminal, making it hard to focus in on the relevant output (or is it just my tired eyes?!) For example, this is what it is now:

BEFORE

Could be something like:

AFTER

- Caveat: This will depends on how you are interacting with python. So I guess other people's outputs could look different to mine? However, putting in a space will still make it clearer on any platform, I think.

## Next Steps
- Go through the phys2bids code and locate where there is text that is printed out to screen. Add a tab at the start of each line.

vinferrer:
Closes #196 

## Proposed Changes

  - adds the parameter ``format='%(levelname)s:\t%(message)s')`` to the logger
  - Results:
image

### DIFF
--- a/phys2bids/phys2bids.py
+++ b/phys2bids/phys2bids.py
@@ -169,13 +169,13 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
 
     if quiet:
         logging.basicConfig(level=logging.WARNING,
-                            handlers=[log_handler, sh])
+                            handlers=[log_handler, sh], format='%(levelname)s:\t%(message)s')
     elif debug:
         logging.basicConfig(level=logging.DEBUG,
-                            handlers=[log_handler, sh])
+                            handlers=[log_handler, sh], format='%(levelname)s:\t%(message)s')
     else:
         logging.basicConfig(level=logging.INFO,
-                            handlers=[log_handler, sh])
+                            handlers=[log_handler, sh], format='%(levelname)s:\t%(message)s')
 
     version_number = _version.get_versions()['version']
     LGR.info(f'Currently running phys2bids version {version_number}')


smoia:
Quick question: is there a way to have a different format for different types of loggers? If so, we could get something more similar to what @RayStick was proposing in #196 .

smoia:
(Apparently, there is: [see this Stack Overflow](https://stackoverflow.com/questions/1343227/can-pythons-logging-format-be-modified-depending-on-the-message-log-level))

smoia (changes requested):
This is going to look great once we merge it! However, I'd like to see a different implementation more similar to what @RayStick was thinking about.

RayStick:
Looking good, and thanks for working on it. I agree with Stefano's comments above

vinferrer:
What about this:
image

smoia (approved):
Looks great! Go for it!

RayStick:
Looking great. I will test it locally this morning, and finish the review.

tsalo:
The output looks beautiful. I did notice a single multi-line logged message that somewhat mars the perfection. Is it worth a GFI?

Screen Shot 2020-10-13 at 9 54 33 AM

smoia:
@tsalo can you open (sounds new...) a GFI for this?

RayStick (approved):
Looks much clearer on the terminal! :)

RayStick:
1 of the code coverage checks is failing. Do we need to rectify that before I merge?

eurunuela:
> 1 of the code coverage checks is failing. Do we need to rectify that before I merge?

That's alright. No need to rectify as we're still accomplishing our `>90%`  goal.

smoia:
Mainly, no need to rectify because it's not that @vinferrer is adding untested lines.

I really don't understand this behaviour in codecov - it happens to me all the time too, even if I don't modify the code of the files that are covered, or if I add tests for those. Does anyone know why this happens?

RayStick:
Yes, I was a bit confused by the report when I viewed it!

smoia:
Yes, I can imagine!
As a rule of thumb, when you see that codecov fails, check if the coverage drop is real or not. If the report doesn't show any new line that is not tested, then the drop can be ignored. If the drop is related to the PR adding untested lines of code, you can ask to add new tests or open an issue to create such tests.

PR #313, opened 2020-10-09
Setup auto release workflow

### LINKED ISSUE #239: auto doesn't find the Github token when running in Travis
@hipstersmoothie we may need your help here.

@eurunuela and I merged in the PR to run auto, but at the moment [Travis says it's missing a token](https://travis-ci.org/github/physiopy/phys2bids/jobs/698514747).
I created the token a while ago, so I don't understand why it's not working. Should I create a token for the organisation?

Could you have a look at our configuration?
[This is Travis config](https://github.com/physiopy/phys2bids/blob/master/.travis.yml)
[This is auto](https://github.com/physiopy/phys2bids/blob/master/.autorc).

### LINKED ISSUE #311: strip "v" prefix from phys2bids.__version__, make it semver
ATM
```
$> python -c 'import phys2bids; print(phys2bids.__version__)'
v2.1.0+119.gccd39e8
$> git co v2.1.0
$> python -c 'import phys2bids; print(phys2bids.__version__)'
v2.1.0
```
which is awesome that it includes git information, but 

- not good that has `v` prefix.  `v` in git tags is ok, but undesired for `__version__` which should be just the version .  Check other modules (e.g. numpy etc) -- they all lack any `v` prefix in `__version__`.  Dropping it would make `__version__` immediately a proper semver (although "build" doesn't support sorting, but sorting of git states even with index should not be relied upon anyways): 

```
In [1]: import semver                                                                                                   

In [2]: semver.parse("v2.1.0+119.gccd39e8")                                                                             
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-8d6227dd9bc8> in <module>
----> 1 semver.parse("v2.1.0+119.gccd39e8")

/usr/lib/python3/dist-packages/semver.py in parse(version)
     19     match = _REGEX.match(version)
     20     if match is None:
---> 21         raise ValueError('%s is not valid SemVer string' % version)
     22 
     23     verinfo = match.groupdict()

ValueError: v2.1.0+119.gccd39e8 is not valid SemVer string

In [3]: semver.parse("2.1.0+119.gccd39e8")                                                                              
Out[3]: 
{'major': 2,
 'minor': 1,
 'patch': 0,
 'prerelease': None,
 'build': '119.gccd39e8'}
```

related:  please use annotated git tags (although that would require tagging locally and then minting a release from pushed tag on github).  I typically tag locally, add changelog entry into its annotation, push to github , go to releases, edit release for the tag and do not enter/change anything, just "save" it and then it reuses annotation from the git tag (changelog).   

```shell
(git)lena:~/proj/misc/phys2bids[tags/v2.1.0]git
$> git describe       
fatal: No annotated tags can describe '4e0cf9c8683d11528e01c8e3d7d63e179c6bd6f2'.
However, there were unannotated tags: try --tags.
$> git describe --tags
v2.1.0
```

smoia:
Closes #311 and closes #239

Second attempt to integrate `auto` as a workflow, this time as a GitHub action. IF this works, we have a great automated workflow from PR merge to Pypi package (with the other workflow).

## Proposed Changes

  - Add `.autorc` configuration file
  - Add github workflow that * should * retrieve the latest version of auto (might not be elegant) and then run it 
  - Add a couple of badges.

I can't wait to see if this works. So exciting!

@hipstersmoothie, could you please have a look at this PR?
@yarikoptic and @jwodder, you could have a look into this too - if you implemented a different solution I'd like to compare them, and otherwise if this works it should help you with datalad/datalad#4999 and dandi/dandi-cli#252!

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -1,0 +1,86 @@
+{
+  "plugins": [
+    "git-tag",
+    "all-contributors",
+    "conventional-commits",
+    "first-time-contributor",
+    "released"
+  ],
+  "release": {
+    "prerelease": true
+  },
+  "owner": "physiopy",
+  "repo": "phys2bids",
+  "name": "Stefano Moia",
+  "email": "s.moia@bcbl.eu",
+  "labels": [
+    {
+      "name": "Majormod",
+      "changelogTitle": "💥 Breaking Change",
+      "description": "This PR breaks compatibility, and increments the major version (+1.0.0)",
+      "releaseType": "major",
+      "overwrite": true
+    },
+    {
+      "name": "Minormod",
+      "changelogTitle": "🚀 Enhancement",
+      "description": "This PR generally closes an `Enhancement` issue. It increments the minor version (0.+1.0)",
+      "releaseType": "minor",
+      "overwrite": true
+    },
+    {
+      "name": "Minormod-breaking",
+      "changelogTitle": "💥 Breaking Change during development",
+      "description": "For development only, this PR increments the minor version (0.+1.0) but breaks compatibility",
+      "releaseType": "minor",
+      "overwrite": true
+    },
+    {
+      "name": "BugFIX",
+      "changelogTitle": "🐛 Bug Fix",
+      "description": "This PR generally closes a `Bug` issue, and increments the patch version (0.0.+1)",
+      "releaseType": "patch",
+      "overwrite": true
+    },
+    {
+      "name": "Documentation",
+      "changelogTitle": "📝 Documentation",
+      "description": "This issue or PR is about the documentation ",
+      "releaseType": "none",
+      "overwrite": true
+    },
+    {
+      "name": "Testing",
+      "changelogTitle": "⚠️ Tests",
+      "description": "This is for testing features, writing tests or producing testing code",
+      "releaseType": "none",
+      "overwrite": true
+    },
+    {
+      "name": "Internal",
+      "changelogTitle": "🏠 Internal",
+      "description": "Changes affect the internal API. It doesn't increase the version, but produces a changelog",
+      "releaseType": "none",
+      "overwrite": true
+    },
+    {
+      "name": "Outreach",
+      "changelogTitle": "🖋️  Outreach",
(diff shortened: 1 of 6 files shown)


smoia:
(BTW, I set up the GH_TOKEN as an organisation secret. Is that right?)

eurunuela (commented):
I guess this looks good. I don't think I'm the best person to review this PR though.

.autorc
```diff
@@ -0,0 +1,86 @@
...
+    },
+    {
+      "name": "Minormod",
+      "changelogTitle": "🚀 Enhancement",
+      "description": "This PR generally closes an `Enhancement` issue. It increments the minor version (0.+1.0)",
+      "releaseType": "minor",
+      "overwrite": true
+    },
+    {
+      "name": "Minormod-breaking",
```
tsalo, line 32: Is this still valid when `phys2bids` is past alpha (i.e., 1.0.0+).

tsalo (approved):
My review is basically the same as @eurunuela's. I read through everything and it makes sense, but since I'm not familiar with `auto`, I'm not sure how valuable my review is.

> (BTW, I set up the `GH_TOKEN` as an organisation secret. Is that right?)

That seems reasonable. It looks like you need to add your personal access token to the GitHub repository's secret environmental variables. I'm not sure if organization-level secrets are automatically used for individual repositories, but it makes sense that they would be.

.autorc
smoia: We could remove it, but honestly it' mainly there to maintain the same configuration across `physiopy`. I doubt we'll ever use that label!

smoia:
Schermata da 2020-10-10 19-01-20
This is what I did for the secrets.

eurunuela:
I say we give this a shot if you feel comfortable with it @smoia

smoia:
Ok, let's try this out.

smoia:
:rocket: PR was released in [`2.2.1`](https://github.com/physiopy/phys2bids/releases/tag/2.2.1) :rocket:

PR #327, opened 2020-10-21
Adopt Black and isort

### LINKED ISSUE #324: Going black (i.e. adopting an _opinionated_ python code formatter)
I was already thinking about proposing it, but then @tsalo applied it on his #219 so it's a good moment to propose it.

In the line of automating all that is automatable, there is one more thing we should adopt, a _code formatter_.

Code formatters are scripts that chew our code and our PRs and spit out a standardised formatted version of the code, _as long as the code doesn't contain big mistakes_. They shouldn't change radically the code, rather adjust it with simple rules like "always use `'` instead of `"` for strings" or "always adopt a certain indentation rule in your scripts". "Chill" code formatters just check that a coding style is respected (e.g. `autopep8` checks that `pep8` is respected), while there are "opinionated" code formatters that format the code even if it respect a standard (in our case, `pep8`).

One (apparently) popular opinionated python code formatter is [**black**](https://github.com/psf/black). It's very strict and either we agree with its rules or we don't.
Another opinionated code formatter is [YAPF](https://github.com/google/yapf). It's less strict, in the sense that it allows for some customisations, but it does a similar job.

There are pros and cons in adapting a code formatter.
The biggest pro: our code will always be formatted in the same way, independently from its author, so we can forget comments like "please reformat following our style" in reviews and other minutiae.
Cons: If we ask the commits to be already formatted, contributors needs to run black locally before a PR. Alternatively, if we run it during a PR merge (we need to understand how to do it smartly), the code that we will see in the repo might not be exactly the same that we wrote - although it will be very similar.

@physiopy/all, what do you think about adopting an opinionated code formatter? And what do you think about adopting black?

## Outstanding questions

1. Should we adapt an opinionated code formatter?
2. If so, which one between black and YAPF?

PS: Labelling this urgent because we might need to wait for this before merging #219 - and we need to do that in the upcoming month. Deadline is next dev call (remember that next month it will be during the first week).

tsalo:
Closes #324. 

## Proposed Changes

  - Change `flake8` dependency to `flake8-black`. Calling `flake8` in the CI will use `flake8-black` automatically.
  - Ignore E203 (whitespace before ':'), which `flake8` flags, but which is actually not PEP8 compliant. See [here](https://black.readthedocs.io/en/stable/compatible_configs.html#flake8) for an explanation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,10 +97,8 @@ jobs:
           pkg-manager: pip-dist
       - run:
           name: Check style
-          command: echo "Skipping for now"  # |
-            # isort --check .
-            # black --check phys2bids
-            # flake8 ./phys2bids
+          command: |
+            flake8 ./phys2bids
 
   merge_coverage:
     working_directory: /tmp/src/phys2bids
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,9 @@ jobs:
       - run:
           name: Check style
           command: |
-            flake8 ./phys2bids
+            isort --check .
+            black --check nigsp
+            flake8 ./nigsp
 
   merge_coverage:
     working_directory: /tmp/src/phys2bids
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,8 +99,8 @@ jobs:
           name: Check style
           command: |
             isort --check .
-            black --check nigsp
-            flake8 ./nigsp
+            black --check phys2bids
+            flake8 ./phys2bids
 
   merge_coverage:
     working_directory: /tmp/src/phys2bids
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,14 @@ repos:
     -   id: check-added-large-files
     -   id: check-case-conflict
     -   id: check-merge-conflict
+-   repo: https://github.com/psf/black
+    rev: 22.10.0
+    hooks:
+    -   id: black
+-   repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+    -   id: isort
 -   repo: https://github.com/pycqa/flake8
     rev: 6.0.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
     hooks:
     -   id: black
 -   repo: https://github.com/pycqa/isort
-    rev: 5.10.1
+    rev: 5.12.0
     hooks:
     -   id: isort
 -   repo: https://github.com/pycqa/flake8
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,6 +7,7 @@
 
 import os
 import sys
+
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
--- a/phys2bids/__init__.py
+++ b/phys2bids/__init__.py
@@ -1,4 +1,5 @@
 """Main init for phys2bids package."""
 from ._version import get_versions
-__version__ = get_versions()['version']
+
+__version__ = get_versions()["version"]
 del get_versions
--- a/phys2bids/_version.py
+++ b/phys2bids/_version.py
@@ -1,4 +1,3 @@
-
 # This file helps to compute a version number in source trees obtained from
 # git-archive tarball (such as those provided by githubs download-from-tag
 # feature). Distribution tarballs (built by setup.py sdist) and build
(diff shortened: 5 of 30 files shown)


tsalo:
There are surprisingly few style issues, which is nice to see!

@smoia, do you want to use `isort` as well? There's a `flake8-isort` plugin that could be added to the dependencies, and I think we'd need some `isort`-specific settings in `setup.cfg` to fit with `black`.

eurunuela:
Thank you @tsalo ! It's great to see we were already close to the style required by black.

> EDIT: Given that we don't want to let PRs sprawl, I think it would be good to open a _separate_ PR (to be merged before this one) in which `black` is run on the whole repository. WDYT?
 
I'd open a separate PR.

smoia:
Ok, one quick thought about this.
I thought we were going to add a CI job to run black either during PR merge.

This solution obliges the user to run black locally instead. I don't love the idea, mainly because it complicates (a little, but still complicates) the process of contributing to the repos - it might scare contributors away (especially new ones) or force them to adopt black during coding. Can we check if setting up a CI action (actually, adding a job in the autorelease GH action would be better) is feasible?

tsalo:
I considered using the [black GitHub Action](https://github.com/marketplace/actions/black-code-formatter) in NiMARE, but I was wary to give an Action the ability to directly commit to folks' pull requests. I chose to just use a stricter linter in that situation. It doesn't _require_ contributors to run `black` (they could just fix things manually if they wanted), but it does flag poorly formatted code.

That said, it's up to you. I can try switching things out with an Action, but I honestly think you'd be the best person to set it up. I don't use Actions much (if at all), and the last time I tried it messed up my local repo.

smoia:
I see - and understand the warning you're raising.
This might be worth discussing between developers (next meeting or back in #324).

I don't love to force users to adopt opinionated code formatters (or even worse a very opinionated code style) locally, honestly.
This is because I'd be the first to be pushed away if I had to.
Since I'll hardly adopt double quotes or line splitting during coding, I know I won't adjust things manually. I can run black before readying a PR for review, but if I need to work on the code again after a review I would be disoriented, making the process a bit more bothersome. I might even end up committing, running black and re-commiting. Uff (and I know a couple of physiopy contributors that might be even more exasperated by this).
Besides, new users might already overcome the challenge of understanding the library, (possibly) learn and code in python, install or run an easy-peasy linter (still problematic for some of the more expert contributors), possibly add documentation and tests. Asking to adopt a strict format or run black might be a bit too much. Especially since we're not talking about _unreadable_ vs _readable_ code - IMMO "simple" flake8 is not a poorly formatted code, and going through #328 I find blackened code less clear than the previous version (I know it's a matter of habit).

However, if @physiopy/all prefer to run black locally before PR, we can do that - let's keep the discussion flowing.

eurunuela:
> However, if @physiopy/all prefer to run black locally before PR, we can do that - let's keep the discussion flowing.

I'd rather have the linter tell us what to change. I think asking contributors to run black locally is a bit too much. I think we should make our lives easier and not complicate our contributions too much!

smoia:
What if we discuss this in the upcoming meeting?
 We can start the agenda now, given that it's next week.

tsalo:
Where did we land on this? Should I close?

smoia:
Sorry @tsalo, I've been pretty busy with the rest of my commitments. I need just a bit more time to work on this!

smoia:
Hey @tsalo!
Would you still be interested in following the black-ification of the library?
If so, the library changed a little, so could you rebase on the latest master?

After that, we can implement pre-commits to automatically run isort and black on the files.

tsalo:
I'm happy to give it a try. I'll try to fix up this PR later today.

smoia:
Thank you!

smoia:
Let me know when you want me to review to merge!

BTW, I'm going to configure Pre Commit on all Physiopy packages, so if you want to go ahead and add it here too, I'm following [this configuration](https://github.com/physiopy/physiopy-repository-template/blob/master/.pre-commit-config.yaml) with [this setup](https://github.com/physiopy/physiopy-repository-template/blob/225511950c56fc343d6bf9106e247babf63bf19a/setup.cfg#L71) from our [repository template](https://github.com/physiopy/physiopy-repository-template)

tsalo:
@smoia I added the pre-commit config, but I've never used pre-commit before. Is adding the config file enough?

smoia:
Hey @tsalo! On your side it should be sufficient, although you may want to [activate it locally](https://pre-commit.com/#quick-start). On my side, I just allowed pre-commit to run on phys2bids.

pre-commit.ci run

smoia:
Oh, and we're not doing checks on tests, at least for docstrings.

smoia:
@tsalo I tried to update this PR in order to merge it, but I get an error I can't figure out on the style check.

Locally, isort and black skip all files. Remotely, something isn't happy.

Do you have ideas on why?

smoia:
Ok, problem found. It was "just" a black version issue.

smoia (approved):

PR #344, opened 2020-11-12
txt.py and acq.py reorganization into io.py

### LINKED ISSUE #295: Reorganize I/O functions
## Summary

I find the `phys2bids.interfaces.[modality].populate_phys_input()` structure unnecessarily confusing. Why not have `phys2bids.io.load_matlab()`, `phys2bids.io.load_text()`, and `phys2bids.io.load_acq()`?

Note: this stems from https://github.com/physiopy/phys2bids/pull/292#issuecomment-676525617.

## Additional detail

I understand the potential benefits to having multiple functions with the same name (e.g., if you wanted to use `eval` and such to call functions), although, even in those cases, I don't know if it's worth the confusion. In `phys2bids`' case, however, the functions are called directly in the relevant code, so there isn't much benefit to this structure.

Also, `populate_phys_input()` as a function name mostly makes sense if you have multiple classes you might want to read in from files. In `phys2bids`' case, though, you really just want to read them into `BlueprintInput` objects (e.g., you're not going to read an acq file into a `BlueprintOutput` object), so I don't think there's any need to be so specific about the function name. A generic `load_` or `read_` should be sufficient.

## Next Steps

* Move the relevant functions to the same submodule (`phys2bids.io`). If you're concerned about their respective imports, you can just move the tool-specific imports into the function definitions.
* Rename the functions from `populate_phys_info()` to `load_matlab()`, `load_acq()`, and `load_text()` (or similar).
* Update tests and API documentation.

vinferrer:
Closes #295 

Merges `txt.py`and `acq.py` into `io.py`

## Proposed Changes

* Move the relevant functions to the same submodule (`phys2bids.io`). If you're concerned about their respective imports, you can just move the tool-specific imports into the function definitions.
* Rename the functions from `populate_phys_info()` to  `load_acq_ext()`, and `load_text_ext()` (or similar).
* Update tests and API documentation.

## Change Type

- [ ] `bugfix` (+0.0.1)
- [ ] `minor` (+0.1.0)
- [ ] `major`  (+1.0.0)
- [ ] `internal` (no version update)

## Checklist before review

- [x] Please, comment on my PR while it's a draft and give me feedback on the development!


eurunuela:
Thank you @vinferrer ! I don't have the bandwidth right now to review this but I will. 

I see you selected both `minormod` and `majormod`. How's that?

Also, we would be going under 90% code coverage, so we'll need some tests.

vinferrer:
> Thank you @vinferrer ! I don't have the bandwidth right now to review this but I will.
> 
> I see you selected both `minormod` and `majormod`. How's that?
I don't know what to do about it.
> Also, we would be going under 90% code coverage, so we'll need some tests.
That's weird i didn't change any tests i just moved the functions around and modified the imports in the testing

vinferrer:
I forgot to add the testing function, :laughing:

eurunuela:
Now that I think of it, I believe this is an internal update. Not a `majormod` or a `minormod`. It really doesn't change the output of the package but makes our lives easier.

What do you think @tsalo @smoia ?

smoia:
Hey y'all! Sorry, I noticed I didn't add the refactoring to the possible change types. But @vinferrer labelled it right: it's a refactoring.

"Internal" is for the infrastructure. Maybe we can change the label, now that I think about it...

smoia:
BTW: I would get rid of the "interfaces" folder and rename `io.py` into `interfaces.py`. (It's not really an output function, is it?)

tsalo:
An interface implies a class to me, with bidirectional interaction with the source. Having an input/output module also lets you put the `write_json`, `write_file`, etc. functions in the same place.

eurunuela:
I like `io.py` better too.

vinferrer:
so should i get rid of the folder then ?

eurunuela:
> so should i get rid of the folder then ?

I would say yes. It does not make much sense to have the folder with just one file. And usually, `io.py` goes on the same level as `utils.py` for instance.

vinferrer:
okay

vinferrer:
I think this is ready for review then

smoia (commented):
Generally LGTM, but unless everyone else is ok with it, could you remove `_ext` from the name of the functions?

vinferrer:
> Generally LGTM, but unless everyone else is ok with it, could you remove `_ext` from the name of the functions?

Actually no, this is something that i put to remark that we are processing `extensions`  `.txt` is an extension, then we have formats like `acq` that can be in `.txt` or in `.acq` extensions

vinferrer:
This is actually related to the next PR I talk about in #295. Because my goal is to have extension functions that will call in the end to the same process fucntion. This way we will reduce the amount of code

smoia (approved):
Alright, sounds good!
It's ready!

phys2bids/io.py
```diff
@@ -1,18 +1,21 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
-"""phys2bids interface for txt files."""
-
+"""phys2bids interface for acqknowledge files."""
```
eurunuela, line 4: Why do you say it's the interface for acqknowledge files? We read txt files too.

eurunuela (approved):
LGTM! I just have a tiny comment.

phys2bids/io.py
vinferrer: resolved

smoia:
@vinferrer , you merged a PR without being the main reviewer (not a big deal), but also without passing the style check.
Please open a PR now to pass it before anyone else opens a new PR.

smoia:
:rocket: PR was released in [`2.3.0`](https://github.com/physiopy/phys2bids/releases/tag/2.3.0) :rocket:

PR #347, opened 2020-11-13
Add more "change type" options

smoia:
Closes #

## Proposed Changes

  - Add more change type options in the PR template
  -
  -

## Change Type

- [ ] `bugfix` (+0.0.1)
- [ ] `minor` (+0.1.0)
- [ ] `major`  (+1.0.0)
- [ ] `internal` (no version update)
- [ ] `documentation` (no version update)

## Checklist before review

- [ ] I added everything I wanted to add to this PR.
- [ ] \[Code or tests only\] I wrote/updated the necessary docstrings.
- [ ] \[Code or tests only\] I ran and passed tests locally.
- [ ] \[Documentation only\] I built the docs locally.
- [ ] My contribution is harmonious with the rest of the code: I'm not introducing repetitions.
- [ ] My code respects the adopted style, especially linting conventions.
- [ ] The title of this PR is explanatory on its own, enough to be understood as part of a changelog.
- [ ] I added or indicated the right labels.

- [ ] I added information regarding the timeline of completion for this PR.

- [ ] Please, comment on my PR while it's a draft and give me feedback on the development!

PR #349, opened 2020-11-13
Refactor `io` functions related to txt file imports.

vinferrer:
Closes nothing

## Proposed Changes

  - Reducing code duplicity in `io.py`
  - merge `process_acq` and `process_txt` into `process_blueprint_items`
  - Create `extract_header_items` function that gets the header and extracts `interval, orig_units, orig_names` and gives them to `process_blueprint_items`

## Change Type

- [ ] `bugfix` (+0.0.1)
- [ ] `minor` (+0.1.0)
- [ ] `major`  (+1.0.0)
- [x] `refactoring` (no version update)
- [ ] `other`

## Checklist before review

- [x] Please, comment on my PR while it's a draft and give me feedback on the development!


vinferrer:
Well there goes 50 lines of duplicated code

vinferrer:
This is ready for review @smoia, @eurunuela. Once this is merged i will adress the matlab extension update

phys2bids/io.py
```diff
@@ -69,18 +68,22 @@ def check_multifreq(timeseries, freq, start=0, leftout=0):
     return mfreq
 
 
-def process_labchart(channel_list, chtrig, header=[]):
+def process_blueprint_items(channel_list, chtrig, interval, orig_units, orig_names):
     """
-    Process labchart header and channel_list and make a physio_obj.BlueprintInput.
+    Process items header items and channel_list and make a physio_obj.BlueprintInput.
```
eurunuela, line 73: I'm not sure what you mean by `items header items and channel_list`

phys2bids/io.py
```diff
@@ -90,116 +93,26 @@ def process_labchart(channel_list, chtrig, header=[]):
...
-    interval = header[1][0].split()
-    # check the interval is in some of the correct AcqKnowledge units
-    if interval[-1].split('/')[0] not in ['min', 'sec', 'µsec', 'msec', 'MHz', 'kHz', 'Hz']:
+    if interval[-1] not in ['min', 'sec', 'µsec', 'msec', 'MHz', 'kHz', 'Hz', 'hr', 'min', 's',
+                            'ms', 'µs']:
         raise AttributeError(f'Interval unit "{interval[-1]}" is not in a '
-                             'valid AcqKnowledge format time unit, this probably'
-                             'means your file is not in min, sec, msec, µsec, Mhz, KHz or Hz')
-    interval[-1] = interval[-1].split('/')[0]
+                             'valid format frequency or time unit, this probably'
```
eurunuela, line 111:
```suggestion
                             'valid frequency or time unit format, this probably'
```

eurunuela (changes requested):
This is great, thank you @vinferrer ! 

I think some rephrasing of comments must be done first though.

phys2bids/io.py
```diff
@@ -292,13 +214,67 @@ def read_header_and_channels(filename):
...
+    Parameters
+    ----------
+    channel_list: list
+        list with channels. For one of the formats it contains header information
+    header: list
+        list with that contains file header
+
+    Returns
+    -------
+    interval: list
```
tsalo, line 230: `interval : list of X`

numpydoc puts a space on both sides of the `:`. I'd also have a bit more info about the actual contents of the list. list of str, list of float, etc.

phys2bids/io.py
```diff
@@ -292,13 +214,67 @@ def read_header_and_channels(filename):
...
+        contains interval or frequency value and units
```
tsalo, line 231: This description confuses me a bit. It has the interval _or_ frequency value _and_ units. So is it a list of dictionaries? Or tuples?

I also don't think there's any need to say "contains". I'd say something like "interval or frequency value for each channel."

phys2bids/io.py
```diff
@@ -69,18 +68,22 @@ def check_multifreq(timeseries, freq, start=0, leftout=0):
     return mfreq
 
 
-def process_labchart(channel_list, chtrig, header=[]):
+def process_blueprint_items(channel_list, chtrig, interval, orig_units, orig_names):
```
tsalo, line 71: Since this function (1) does some processing on inputs and (2) generates a BlueprintInput object, why not just `generate_blueprint()`?

phys2bids/io.py
```diff
@@ -292,13 +214,67 @@ def read_header_and_channels(filename):
...
+            orig_units.append(header[index1 + 1][0])
+    else:
+        raise AttributeError('This file format is not supported yet for txt files')
+    return interval, orig_units, orig_names
+
 
-    Extract header and deduce from it the format file,
-    afterwards pass the needed information to
-    the corresponding reading function.
+def load_txt_ext(filename, chtrig=0):
```
tsalo, line 275: I'd say just `load_txt()`. Including `ext` is somewhat redundant.

phys2bids/io.py
```diff
@@ -292,13 +214,67 @@ def read_header_and_channels(filename):
...
+    """
+    Get txt extension file and processes with different subfunctions.
```
tsalo, line 277: "Read .txt file into a BlueprintInput object.", maybe.

phys2bids/io.py
vinferrer: so channel_list is the first parameter of the function and the header items are the rest of them

phys2bids/io.py
vinferrer:
```
    interval : list of strings
        maximun sampling frequency or interval value and unit for the recording
```
Better? it only contains the value and the unit of one frequency or interval

phys2bids/io.py
vinferrer: Okay but then also ``load_acq_ext`` should be renamed, the thing is that acq is both an extension and a format (or package)

phys2bids/io.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
```
tsalo, line 2: Just noticed this. There's no need for a shebang for files in the package, unless they will be run from the command line (and even then, if they're set up as CLIs there's probably no need). See [here](https://stackoverflow.com/a/19305076/2589328).

phys2bids/io.py
```diff
@@ -69,18 +68,22 @@ def check_multifreq(timeseries, freq, start=0, leftout=0):
...
     ----------
-    channel_list: list
+    channel_list : list of strings
         list with channels only
     chtrig : int
         index of trigger channel, starting in 1 for human readability
-    header: list
-        list with that contains file header
+    interval : list of strings
+        maximum sampling frequency or interval value and unit for the recording
```
tsalo, line 82: So this might be something like `["400Hz", "300Hz"]`?

phys2bids/io.py
```diff
@@ -292,27 +214,81 @@ def read_header_and_channels(filename):
...
+            # since units are in the line imediately after we get the units at the same time
+            orig_units.append(header[index1 + 1][0])
+    else:
+        raise AttributeError('This file format is not supported yet for txt files')
+    return interval, orig_units, orig_names
+
+
+def load_txt(filename, chtrig=0):
+    """
+    Read .txt file into a BlueprintInput object.
```
tsalo, line 277: Worth noting in either the single-line description or the multi-line description that this can read in both acqknowledge and labchart files that have been exported to txt format.

phys2bids/io.py
smoia: It's true, it's not necessary. Since it doesn't hurt anyone though (and AFAIK it's not against pep8), let's leave it there - it helps (some) editors and (some) humans to know it's python. (You might argue the extension should be enough, but still.)

phys2bids/io.py
eurunuela: True. We should definitely mention that.

phys2bids/io.py
vinferrer: nope, more something like ``["400", "Hz"]`` or ``["1", "ms"]``  that's why i say for the recording, because in txt files the only information we have about frequency is the maximum. That's why we have the ``check_frequency`` function, to get the frequency for each channel

phys2bids/io.py
```diff
@@ -209,46 +122,55 @@ def process_acq(channel_list, chtrig, header=[]):
...
+            if interval[-1] == 'hr':
+                interval[0] = float(interval[0]) * 3600
+                interval[-1] = 's'
+            elif interval[-1] == 'ms':
+                interval[0] = float(interval[0]) / 1000
+                interval[-1] = 's'
+            elif interval[-1] == 'µs':
+                interval[0] = float(interval[0]) / 1000000
+                interval[-1] = 's'
+            elif interval[-1] == 'sec':
```
smoia, line 149: I'm not totally sure that this test is necessary any more, given that there is no use for `interval[-1]` after here. In fact, you could even remove all the `interval[-1] = 's' in the elif statements!

phys2bids/io.py
```diff
@@ -209,46 +122,55 @@ def process_acq(channel_list, chtrig, header=[]):
...
                 interval[-1] = 's'
-            elif interval[-1].split('/')[0] == 'msec':
+            elif interval[-1] == 'msec':
                 interval[0] = float(interval[0]) / 1000
                 interval[-1] = 's'
-            elif interval[-1].split('/')[0] == 'µsec':
+            elif interval[-1] == 'µsec':
                 interval[0] = float(interval[0]) / 1000000
                 interval[-1] = 's'
+            if interval[-1] == 'hr':
```
smoia, line 140:
```suggestion
           elif interval[-1] == 'hr':
```
I know it's not super mandatory, it's just to have it in the same statement.

phys2bids/io.py
```diff
@@ -209,46 +122,55 @@ def process_acq(channel_list, chtrig, header=[]):
...
     units = ['s', ]
     units = units + orig_units
-    # get channels
+    timeseries = list(map(list, zip(*channel_list)))
+    freq = [1 / interval[0]] * len(timeseries)
     timeseries = [np.array(darray) for darray in timeseries]
-    duration = (timeseries[0].shape[0] + 1) * interval[0]
-    t_ch = np.ogrid[0:duration:interval[0]][:-1]  # create time channel
-    timeseries = [t_ch, ] + timeseries
+    # check the file has a time channel if not create it and add it
```
smoia, line 165:
```suggestion
    # Check if the file has a time channel, otherwise create it.
```

phys2bids/io.py
```diff
@@ -209,46 +122,55 @@ def process_acq(channel_list, chtrig, header=[]):
...
+    # As the "time" doesn't have a column header, if the number of header names
+    # is less than the number of timesieries, then "time" is column 0...
```
smoia, line 167:
```suggestion
    # is less than the number of timeseries, then "time" is column 0...
```

phys2bids/io.py
```diff
@@ -292,27 +214,81 @@ def read_header_and_channels(filename):
     return header, channel_list
 
 
-def load_txt_ext(filename, chtrig=0):
+def extract_header_items(channel_list, header=[]):
     """
-    Populate object phys_input.
+    Process items header and channel_list depending on the format (acqKnowledge and labchart).
```
smoia, line 219:
```suggestion
    Process items header and channel_list depending on the format (AcqKnowledge and labchart).
```

phys2bids/io.py
smoia: Maybe an example helps - can you add it?

phys2bids/io.py
```diff
@@ -90,116 +93,26 @@ def process_labchart(channel_list, chtrig, header=[]):
     ------
     ValueError
         If len(header) == 0 and therefore there is no header
```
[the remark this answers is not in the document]
smoia: No, wait, I see. I think you just need to remove this line.
```suggestion

```

phys2bids/io.py
```diff
@@ -292,27 +214,81 @@ def read_header_and_channels(filename):
...
+    Process items header and channel_list depending on the format (acqKnowledge and labchart).
 
-    Extract header and deduce from it the format file,
-    afterwards pass the needed information to
-    the corresponding reading function.
+    Parameters
+    ----------
+    channel_list : list of strings
+        The channels of the recording
+    header: list
```
smoia, line 225:
```suggestion
    header : list
```

phys2bids/io.py
```diff
@@ -292,27 +214,81 @@ def read_header_and_channels(filename):
...
+    interval : list of strings
+        maximun sampling frequency or interval value and unit for the recording
+    orig_units : list of strings
+        contains original channels units
+    orig_names : list of strings
+        contains original channels name
+
+    Raises
+    ------
+    ValueError
```
smoia, line 239: Same as before. Idk which one is better (probably Attribute?), but can you correct it?

smoia: Maybe it's actually two different types of error instead?

phys2bids/io.py
```diff
@@ -292,27 +214,81 @@ def read_header_and_channels(filename):
...
+            # since units are in the line imediately after we get the units at the same time
+            orig_units.append(header[index1 + 1][0])
+    else:
+        raise AttributeError('This file format is not supported yet for txt files')
+    return interval, orig_units, orig_names
+
+
+def load_txt(filename, chtrig=0):
+    """
+    Read  acqknowledge and labchart files in .txt format into a BlueprintInput object.
```
smoia, line 277:
```suggestion
    Read AcqKnowledge and labchart files in .txt format into a BlueprintInput object.
```

phys2bids/io.py
```diff
@@ -292,27 +214,81 @@ def read_header_and_channels(filename):
...
     chtrig : int
         index of trigger channel, starting in 1 for human readability
 
     Returns
     -------
     phys_in
-        Raises
-    ------
 
+    Raises
```
smoia, line 290: You can safely remove this section. It's true that the functions called here raise such errors, but not this one in particular!

phys2bids/io.py
```diff
@@ -292,27 +214,81 @@ def read_header_and_channels(filename):
...
     ----------
-    filename: str
+    filename : str
         path to the txt Labchart file
     chtrig : int
         index of trigger channel, starting in 1 for human readability
 
     Returns
     -------
     phys_in
```
smoia, line 288:
```suggestion
    phys_in
        BlueprintInput object
```

phys2bids/io.py
```diff
@@ -321,29 +297,19 @@ def load_txt_ext(filename, chtrig=0):
...
     """
     Populate object phys_input from acq extension files.
 
     Parameters
     ----------
-    filename: str
+    filename : str
         path to the txt labchart file
     chtrig : int, optional
           index of trigger channel. Default is 0.
```
smoia, line 315:
```suggestion
        index of trigger channel. Default is 0.
```

phys2bids/io.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
+"""phys2bids interfaces for txt  and acq extension files."""
 
 import logging
 from collections import Counter
```
smoia, line 6: I can't comment at the right line (again), but here's a little thing that might have been overlooked in #347.
Line 11: `bioread` is not a mandatory dependency (nor it should be) - it's only used if acq file have to be processed.
Before, when each interface had its own script, it was not a problem because the acq interface would have been called ONLY if an acq file was going to be run. Now that the interfaces are going to be initialised anyway, if this import is left here, the program will try to import `bioread` even if a txt file is read - and if the user doesn't have it installed (given that it's not a mandatory library), it will raise an error.
Two solutions for this: either you move this import to the first line in `load_acq`, or you `try` this import and raise an error in load_acq if it's missing. I suggest to go for the first solution - makes the code easier.

smoia: BTW, this will be the same for matlab related libraries.

phys2bids/phys2bids.py
```diff
@@ -219,11 +219,11 @@ def phys2bids(filename, info=False, indir='.', outdir='.', heur_file=None,
     # Read file!
     LGR.info(f'Reading the file {infile}')
     if ftype == 'acq':
-        from phys2bids.io import load_acq_ext
-        phys_in = load_acq_ext(infile, chtrig)
+        from phys2bids.io import load_acq
+        phys_in = load_acq(infile, chtrig)
```
smoia, line 223: See the comment on the import of `biored`. If you move the import internally to this function, then this is a good solution. Otherwise, you can move this import to the beginning of the file (but I suggest not to).

smoia (changes requested):
Nice work! Great refactoring @vinferrer !
There are a couple of things to correct here and there - check the comments.

phys2bids/io.py
eurunuela: Okay, then we should probably make the docstring clearer. It has to be descriptive for future developers too.

phys2bids/io.py
```diff
@@ -90,116 +93,26 @@ def process_labchart(channel_list, chtrig, header=[]):
...
-                             'valid AcqKnowledge format time unit, this probably'
-                             'means your file is not in min, sec, msec, µsec, Mhz, KHz or Hz')
-    interval[-1] = interval[-1].split('/')[0]
+                             'valid frequency or time unit format, this probably'
+                             'means your file is not in min, sec, msec, µsec, hr, min, s, ms, µs,'
+                             'Mhz, KHz or Hz')
     # Check if the header is in frequency or sampling interval
-    if 'Hz' in interval[-1].split('/')[0]:
+    if 'Hz' in interval[-1]:
         print('frequency is given in the header, calculating sample Interval'
```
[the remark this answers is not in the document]
eurunuela: I also think these changes should be added.

eurunuela (changes requested):
The code looks great, thank you @vinferrer ! However, I think @smoia 's suggestions need to be added as well as my little comment.

It's mainly rephrasing of comments and docstrings.

phys2bids/io.py
vinferrer: It is necessary, remember that ``interval=["400", "Hz"]`` . Therefore we need to access only the second item, which is the one containing the unit information

eurunuela (approved):
I'm happy with the changes. Thank you @vinferrer !

vinferrer:
@smoia @tsalo

phys2bids/io.py
smoia: Let me try again.
The last moment you use `interval[-1]` in this function is during this `if` statement, to see if the value of `interval[0]` should be "corrected". After this `if` statement, `interval[-1]` becomes unused - from what I see, you're adding the unit `s` in the `unit` list in the hard way (coherently with your script).
For this reason, every `interval[-1]='s'` in this `elif` statement can be removed (unless you're keeping it for testing, but then please write it).
Lines 152-154 are actually superfluous. Line 152 and 154 do exactly what `phys2bids.bids.bidsify_unit()` does downstream, and line 153 is the same as line 156. The only change you would have to do (that is not wrong) is adding the option in line 131 for `interval[-1] != 's' or interval[-1] != 'sec'`. Which also makes sense for the message you're giving after in 132. It would be weird, if I know that I set up my labchart to record my sampling interval in seconds, to see that the sampling interval is not second.

You can save almost ten lines, I would do it!

smoia (commented):
One last thing to address (either change or reply to the suggestion), but then it's ready!

vinferrer:
Okay, now I understand. It's done

phys2bids/io.py
```diff
@@ -128,32 +128,22 @@ def generate_blueprint(channel_list, chtrig, interval, orig_units, orig_names):
     else:
         # check if interval is in seconds, if not change the units to seconds and
         # calculate frequency
-        if interval[-1] != 's':
+        if interval[-1] != 's' or interval[-1] != 'sec':
```
tsalo, line 131: probably easier to have `interval[-1] not in ('s', 'sec')`

phys2bids/io.py
```diff
@@ -128,32 +127,22 @@ def generate_blueprint(channel_list, chtrig, interval, orig_units, orig_names):
...
-                interval[-1] = 's'
             elif interval[-1] == 'µs':
                 interval[0] = float(interval[0]) / 1000000
-                interval[-1] = 's'
-            elif interval[-1] == 'sec':
-                interval[0] = float(interval[0])
-                interval[-1] = 's'
-        else:
-            interval[0] = float(interval[0])
+        interval[0] = float(interval[0])
```
smoia, line 145:
```suggestion
    else:
        interval[0] = float(interval[0])
```

smoia (changes requested):
Last little suggestion, then it's going to be good on my side.

smoia (approved):
LGTM!

Great job!

smoia:
(and yeah, we should find a way to avoid codeconv to protest every time)

smoia:
@eurunuela if you're happy you can merge.

eurunuela:
Thank you @vinferrer !

smoia:
:rocket: PR was released in [`2.3.0`](https://github.com/physiopy/phys2bids/releases/tag/2.3.0) :rocket:

PR #350, opened 2020-11-13
Adjusts trigger "guess" to channel mean

### LINKED ISSUE #211: Improve trigger threshold computation
## Detailed Description

While working on #206 @sangfrois and I realised that the current approach to compute the threshold fails for multi-runs recordings. Digging into the matter a bit deeper, I realised that this happens if, in absence of noise:
- The average is different from zero and there are too many peak points, compared to the baseline points.
- The average is different from zero and is considerable compared to the level the peaks get to.

As an easy fix, we could set the threshold to be one standard deviation from the average, instead of two standard deviations from the average.
Testing it a bit more with noisy signal, I realised that one standard deviation from average wouldn't be enough to proper threshold points if:
- the noise is considerable compared to the level the peaks get to
- the noise is not much, but there are not enough peak points
- the noise is not much, but the average is not around zero.

A less easy fix in this case would be reconsidering the options in #173 and implement a different approach.

## Possible Implementation

1. We could simply retune the threshold as one standard deviation from the average of the trigger channel, OR
2. We could implement a different approach, for instance the one based on distributions, OR
3. We could do both, keep the standard deviation as "default" approach, and implement a different approach upon user request. 

In case we decide to go for option 1, please relabel this issue as bug and the PR as BugFIX

62442katieb:
Closes #211 

If trigger channels are roughly binary (i.e., "off" is always the same low number (± error) and "on" is always the same higher number (± error)), then then, mathematically, any number above the mean would be "on" and every number below the mean would be "off". 

## Proposed Changes

  - Changes thresh calculation from `mean + 2 * stdev` to `mean`
  - Includes comments detailing my logic, jic

## Change Type

- [x] `bugfix` (+0.0.1)
- [ ] `minor` (+0.1.0)
- [ ] `major`  (+1.0.0)
- [ ] `refactoring` (no version update)
- [ ] `other`

## Checklist before review

- [ ] I added everything I wanted to add to this PR.
- [ ] \[Code or tests only\] I ran and passed tests locally.

- [ ] Please, comment on my PR while it's a draft and give me feedback on the development!


phys2bids/physio_obj.py
```diff
@@ -469,7 +469,12 @@ def check_trigger_amount(self, thr=None, num_timepoints_expected=0, tr=0):
         # Check if thr was given, if not "guess" it.
         flag = 0
         if thr is None:
-            thr = np.mean(trigger) + 2 * np.std(trigger)
+            # If trigger channels are binary 
+            # (i.e., "on" is a higher value and "off" is a lower value)
+            # and each "on" and "off" are each always approzimately the same value
+            # then any value above the mean is "on" and every value below the mean
+            # is "off". 
+            thr = np.mean(trigger)
```
vinferrer, line 477: I see your point, although it happens that the trigger values oscillate in the off and on states so there is no plateau. However,  If you think that will not affect the detection, go for it

62442katieb: My thinking is that the magnitude of the oscillations about "on" and "off" are unlikely to be so large that they approach the mean. Would be happy to try this out with a few different datasets to test that theory!

vinferrer: You are probably right

smoia:
Hi @62442katieb !
It looks like #243 might take a little longer to be merged. How bad would it be for you to cherry pick your changes related to this PR only?

62442katieb:
Hi @smoia ! I'll see if I can't get that done this week so that we can merge this in ASAP.

vinferrer:
Could you merge with main, afterwards i think we can give the OK to this

vinferrer:
Tests are failing what is happening here ?

62442katieb:
> Tests are failing what is happening here ?

What's happening is a big mess as I was working on disentangling #243 related commits from the incredibly simple trigger guess change (which is not looking all that possible at the moment).

smoia:
Ok - let's pause this PR until we merge #243. What do you think @62442katieb ?

62442katieb:
> Ok - let's pause this PR until we merge #243. What do you think @62442katieb ?

That or I can close this PR and open a new, clean one that's not a hot mess.

smoia:
> > Ok - let's pause this PR until we merge #243. What do you think @62442katieb ?
> 
> That or I can close this PR and open a new, clean one that's not a hot mess.

That would be better, if you can!

62442katieb:
Closing and trying again.

Issue #405, opened 2021-06-17
Implement new interfaces for other non-MRI recorded file formats


smoia:
During OHBMH @htwangtw showed me [neo](https://github.com/NeuralEnsemble/python-neo/tree/master), a python package that can read a lot of different physiological file formats (see the list [here](https://neo.readthedocs.io/en/stable/io.html#module-neo.io)), like [Spike2](http://ced.co.uk/products/spkovin) as implemented [here](https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/spike2rawio.py).
@htwangtw also has some code that we could use to work on a phys2bids interface for spike2 [here](https://github.com/htwangtw/adie_ongoingthoughts/blob/main/adie/spike.py).

We could use it to expand our file support to other file formats. Granted, we also would need some actual recordings to work with, but we could start looking around for them.

Some of the file formats are for neurophysiology, some others are for general biosignal - and we should concentrate on those!

htwangtw:
A few things I found while working on this issue:

For Spike2 file, CED has two kinds of files, `.smr` and `.smrx`. The latter is a 64-bits version and going to replace to former. It will be a few release cycle before `.smrx` is supported. (See this [issue](https://github.com/NeuralEnsemble/python-neo/issues/424))
Another news is CED released a [pip-installable API](https://pypi.org/project/sonpy/)

We can either -
1. Limit the support `.smr` and wait for the upstream dependency to update `smrx` support
2. Use sonpy API 

WDYT?

smoia:
Sorry for the late reply - I didn't see this comment!

I'd start with `.smr` support through `neo` (also because I can't find the documentation of `sonpy`), and add `.smrx` support in the future. I wouldn't drop `.smr` support unless we're 100% sure that it's not used any more by the users.
If `sonpy` supports both formats, we could use that API instead.

Does it make sense?

htwangtw:
I had a closer look - `neo` is using `sonpy` under the hood. 
The part of `neo` that broke the loading function of `.smrx` is just missing the header loading info in the 64-bit format. 
I can attempt to fix the upstream dependency so we can just use `neo`

smoia:
Oh wow, that sounds great!

Or we can also use sonpy straight away and get the interface rolling, as you prefer!

smoia:
(as in: I completely trust your judgement, go for the option that is easier or looks more proper for you to implement)

htwangtw:
[...]

PR #410, opened 2021-07-21
Add support for Spike2 files

htwangtw:
Addresses #405.

The current implementation uses the official API, `sonpy`, from CED. `sonpy` is closed sourced.
**In order to support `sonpy`, the minimal python version should be 3.7.** Hence I will mark this as part of the major release.

The scanner volume trigger in the test file was stored as a "marker" channel. The `sonpy` API returns time points of markers, rather than a timeseries. The module I wrote has been tested locally. I cannot run the full test due to bandwidth and data limit. Will check if this breaks other tests.

## Proposed Changes

  - Add `phys2bids.io.load_smr`

## Change Type

- [ ] `bugfix` (+0.0.1)
- [x] `minor` (+0.1.0)
- [ ] `major`  (+1.0.0)
- [ ] `refactoring` (no version update)
- [ ] `other`

## Checklist before review

- [ ] I added or indicated the right labels.

- [ ] Please, comment on my PR while it's a draft and give me feedback on the development!

### DIFF
--- a/phys2bids/io.py
+++ b/phys2bids/io.py
@@ -575,7 +575,7 @@ def load_smr(filename, chtrig=0):
         sonpy.lib.DataType.AdcMark: sonpy.lib.SonFile.ReadWaveMarks,
         sonpy.lib.DataType.RealMark: sonpy.lib.SonFile.ReadRealMarks,
         sonpy.lib.DataType.TextMark: sonpy.lib.SonFile.ReadTextMarks,
-        sonpy.lib.DataType.RealWave: sonpy.lib.SonFile.ReadFloats
+        sonpy.lib.DataType.RealWave: sonpy.lib.SonFile.ReadFloats,
     }
 
     smrfile = sonpy.lib.SonFile(filename, True)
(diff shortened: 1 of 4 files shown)


htwangtw:
Looks like conda doesn't support the latest version of sonpy - I am not familiar with Cirrus CI, if anyone can look into how to get pip to install certain libraries, that would be grand.

smoia (approved):
LGTM (let's wait for the tests)!

htwangtw:
hmmm all python 3.6 checks failed. Looks like a sonpy problem - they use the versioning numbers in a very special way (1.6.x -> python 3.6 support; 1.7.x -> python 3.7 etc)

smoia:
More than that, I believe that sonpy only supports python 3.7 and above (If I read it
Python 3.6 reaches its EOL in a month, so I see two options here:
- we add an exception to not run this test when running python 3.6, or
- we can speed up things and end support for python 3.6 earlier on, removing its tests alltogether.

WDYT?

htwangtw:
Ah good to know! 
I would prefer 
1. explicitly exclude 3.6 for sonpy related functions
2. end support for 3.6 in future iteration. 

Dealing with deprecation and adding a new function in one PR might be too much

smoia:
Agreed! Let's exclude 3.6 testing for sonpy.
Do you need help with it?

htwangtw:
I will have a look and likely working on this next week

smoia:
@htwangtw I tried to update this PR to merge it in.
At the moment, I still cannot figure out how to fix the installation of sonpy - mainly due to the fact that python 3.10 cannot install it.

As a temporary solution, I marked as xfail the sonpy test. I don't love this, but I don't know how to solve it at the moment.

If you agree with the changes, we merge!

smoia:
:rocket: PR was released in [`2.9.0`](https://github.com/physiopy/phys2bids/releases/tag/2.9.0) :rocket:

PR #488, opened 2026-02-23
Improve explanation of pad option in CLI help

### LINKED ISSUE #483: Improve explanation of "pad" option in phys2bids' help
## Expected Behavior
```
  -pad PAD, --padding PAD
                        Padding in seconds used around a single run when separating multi-run session files. Default is 9 seconds.
```

## Actual Behavior

Better to explain the option at more length: `pad` is the amount of seconds of recording that is saved around the real take, both before and after.
At the moment it should be less than the time between two takes (see issue )

## Possible solution

Change description of `pad` in the argparse portion of `cli.py`

BitForge95:
Closes #483 

**What this PR does:**
Updates the argparse help description for the `-pad` / `--padding` argument in `phys2bids/cli/run.py` to better explain that padding is applied both before and after a take, as requested in the issue. 

**Verification:**
- [x] Verified local environment setup.
- [x] Verified `phys2bids --help` outputs the updated description correctly.
- [x] Ran `pytest` locally to ensure no existing tests were broken.

### DIFF
--- a/phys2bids/cli/run.py
+++ b/phys2bids/cli/run.py
@@ -27,8 +27,7 @@ def _get_parser():
         "--input-file",
         dest="filename",
         type=str,
-        help="The name of the file containing physiological "
-        "data, with or without extension.",
+        help="The name of the file containing physiological " "data, with or without extension.",
         required=True,
     )
     optional.add_argument(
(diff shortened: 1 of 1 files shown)


smoia (approved):

smoia:
@all-contributors add BitForge95 for documentation

